PR #24, opened 2015-11-25
Fix logging of unsued options always showing null.

### LINKED ISSUE #15: KAFKA-2026 - Logging of unused options always shows null for the value and is misleading if the option is used by serializers
Our AbstractConfig is also affected by this issue and we need to port whatever patch ends up being used to resolve it. https://issues.apache.org/jira/browse/KAFKA-2026

ewencp:
Fixes #15.


ewencp:
@miguno Patch to resolve #15 that you asked about. This is a direct port of this Kafka patch https://github.com/apache/kafka/commit/23a36eb54958485d1e84c64cc739ca268e96b181 for KAFKA-2026.

miguno:
LGTM

PR #32, opened 2016-07-19
add originalsWithPrefix just like kafka

norwood:
@ewencp


ewencp:
@norwood The addition looks fine. LGTM as long as we've checked that there aren't any other places `RecordingMap` should be used (there are more in Kafka, but not sure if they are all mirrored here).

norwood:
we don't have any of the other methods that use it in kafkaland. i can add those methods if you'd like just for completeness sake

ewencp:
@norwood nah, it's fine. we have a pile of missing features. if we want to do a full sync at some point we should just do the full copy/paste and adjust package names.

PR #33, opened 2016-07-20
turns out if you want to override values, you need values. not originals

norwood:
@ewencp


ewencp:
@norwood Not sure I understand the need for this, but the code LGTM

norwood:
i have a config in which i am creating overrides for a downstream config. so like `blah.downstream.important.config`. `originalsWithPrefix` will only pick up values parsed on config creation, not any defaults in the config. `valuesWithPrefix` will give me the post-parsing config to use so i can have any user provided overrides and my defaults

PR #35, opened 2016-08-24
Remove unnecessary slf4j-log4j dependency that causes confusing warnings.

ewencp:


ewencp:
Not sure who to tag here, but I think @miguno @ijuma @norwood @aayars @gwenshap may have seen this, questions on the mailing list, or want to see this now. We were pulling in slf4j-log4j incorrectly, partly our own fault and partly due to zkclient. This should clean it up, and I think we have more downstream repos to fix up. This should be the source of the annoying "class path contains multiple slf4j bindings" we see pretty much everywhere.

gwenshap:
LGTM. Thank you so much for cleaning this up :)

ewencp:
@gwenshap It was self interest, of course -- I got tired of answer mailing list questions about it :)

ijuma:
@ewencp, thanks for the PR. I had the _pleasure_ of looking at this recently in the context of Streams. ZkClient has a hard dependency on log4j and it doesn't use slf4j (the latter comes from transitively via zookeeper). So, if log4j is not in the classpath at all, ZkClient usage fails with a CNFE. We had to add back a log4j dependency on streams for that reason. Not sure if it's an issue here, but thought I'd mention as I see a log4j exclusion on the zkclient dependency in one of the POMs.

ewencp:
@ijuma Thanks for the heads up. It looks like there are a number of PRs for zkclient trying to get them to remove the log4j dependency and use slf4j instead.

In this particular case, maybe the better change is to just remove our ZkUtils copy anyway. I think people depend on the version in Kafka anyway -- at least it looks like schema-registry does, so I'm not sure what value we're getting by keeping it here.

In any case, all our services ultimately use log4j anyway, so excluding it here *should* just remove the extra copies, unless I'm missing something.

PR #39, opened 2016-11-29
Update ZkClient to 0.10

ewencp:


ewencp:
@ijuma Given that we switched to this in Kafka to get rid of the log4j dependency nonsense, it probably makes sense to do the same for the rest of the platform. Note that I still needed to keep the filters in for log4j -- it seems like it's still being pulled into the compile phase anyway via slf4j-log4j somehow.

ijuma:
LGTM. Regarding the filters, slf4j-log4j is probably being pulled via the transitive `zookeeper` dependency. That's why we also have some exclusions in streams:

https://github.com/apache/kafka/blob/trunk/build.gradle#L700

For `kafka-core`, we don't exclude it because we basically choose the log4j bridge for the Kafka server (it would have been nicer if we only did that for the tarballs instead of the jar, but that's a topic for another day).

ewencp:
Ah, makes sense. Would be nice if zookeeper wouldn't do that...

ijuma:
Yeah, I agree that it would be nice if zookeeper didn't do that. It's a similar situation to kafka core though, they want log4j for the case when it's used a server instead of library.

PR #40, opened 2017-01-26
min needs to start at max_value. hard to get lower than min

norwood:
@nehanarkhede @ewencp


ewencp:
LGTM (looks greatest to me)

🔥 👇

PR #41, opened 2017-02-03
CPKAFKA-437: add ConfigUtils#translateDeprecated and ConfigUtilsTest

cmccabe:
See https://confluentinc.atlassian.net/browse/CPKAFKA-437

### DIFF
--- a/config/src/main/java/io/confluent/common/config/ConfigUtils.java
+++ b/config/src/main/java/io/confluent/common/config/ConfigUtils.java
@@ -1,0 +1,64 @@
+/**
+ * Copyright 2017 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+package io.confluent.common.config;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+public class ConfigUtils {
+  private final static Logger log = LoggerFactory.getLogger(ConfigUtils.class);
+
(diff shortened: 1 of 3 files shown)


config/src/main/java/io/confluent/common/config/ConfigUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+public class ConfigUtils {
+  private final static Logger log = LoggerFactory.getLogger(ConfigUtils.class);
+
+  public static Properties translateDeprecated(Properties props, String[][] keyArray) {
```
ewencp: Might want to document the args a bit. Not obvious at first glance how the `String[][] keyArray` is supposed to be used. In fact, I'm surprised this is a `String[][]` instead of some sort of map.

ewencp: I see in the other PR how this allows you to concisely express the lists without having to allocate, e.g., a bunch of lists (and without building the map, which is annoying without a utility). So probably just a comment here would be enough to explain things.

config/src/main/java/io/confluent/common/config/ConfigUtils.java
```diff
@@ -0,0 +1,64 @@
...
+          "these configuration keys are all synonyms for " + target + ".  Please update " +
+          "your configuration to have only " + target + " set.");
+        newProps.setProperty(target, props.getProperty(deprecated.get(0)));
+      } else {
+        log.warn("Configuration key " + deprecated.get(0) + " is deprecated and may be removed " +
+          "in the future.  Please update your configuration to use " + target + " instead.");
+        newProps.setProperty(target, props.getProperty(deprecated.get(0)));
+      }
+    }
+    return newProps;
```
ewencp: Should we be removing the deprecated ones from the result? We shouldn't need it since we've done the mapping. I'm wondering if we'll pass the deprecated (removed) config along and end up seeing it elsewhere like when unused configs are logged.

config/src/main/java/io/confluent/common/config/ConfigUtils.java
cmccabe: Good point.  I added JavaDoc which should make this clearer.

config/src/main/java/io/confluent/common/config/ConfigUtils.java
cmccabe: Good point.  Let's just remove it.

ijuma:
We need this in 3.2.x as well so that we can merge https://github.com/confluentinc/blueway/pull/1100 to 3.2.x.

PR #43, opened 2017-03-02
ST-240 Coverage+findbugs

norwood:
WIP. testing on jenkins

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,25 +79,7 @@
                 <artifactId>slf4j-api</artifactId>
                 <version>${slf4j.version}</version>
             </dependency>
-            <dependency>
-                <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-log4j12</artifactId>
-                <version>${slf4j.version}</version>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
-                <groupId>junit</groupId>
-                <artifactId>junit</artifactId>
-                <version>${junit.version}</version>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
-                <groupId>org.easymock</groupId>
-                <artifactId>easymock</artifactId>
-                <version>${easymock.version}</version>
-                <scope>test</scope>
-            </dependency>
-            <dependency>
+             <dependency>
                 <groupId>org.apache.zookeeper</groupId>
                 <artifactId>zookeeper</artifactId>
                 <version>${zookeeper.version}</version>
@@ -116,6 +98,25 @@
                         <groupId>log4j</groupId>
                     </exclusion>
                 </exclusions>
+            </dependency>
+            
+            <dependency>
+                <groupId>org.slf4j</groupId>
+                <artifactId>slf4j-log4j12</artifactId>
+                <version>${slf4j.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>junit</groupId>
+                <artifactId>junit</artifactId>
+                <version>${junit.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.easymock</groupId>
+                <artifactId>easymock</artifactId>
+                <version>${easymock.version}</version>
+                <scope>test</scope>
             </dependency>
         </dependencies>
     </dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -46,8 +46,11 @@
         <slf4j.version>1.7.21</slf4j.version>
         <zkclient.version>0.10</zkclient.version>
         <junit.version>4.11</junit.version>
-        <easymock.version>3.0</easymock.version>
+        <easymock.version>3.4</easymock.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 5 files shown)


norwood:
retest this please

norwood:
retest this please

norwood:
@ewencp @xvrl @dguy @cmccabe 

operating plan is to

- push this stuff up in to common
- set all downstream repos parent poms to be 
``` 
    <parent>
        <groupId>io.confluent</groupId>
        <artifactId>common</artifactId>
        <version>${version}</version>
    </parent>
```
- update jenkins jobs for all downstream projects
- ratchet down

norwood:
@aayars

pom.xml
```diff
@@ -43,11 +43,17 @@
     </licenses>
 
     <properties>
+        <argLine></argLine>
```
norwood: jacoco `prepare-agent` needs to set this variable. if we don't have a default set then maven gets angry.

norwood:
retest this please

norwood:
retest this please

norwood:
retest this please

dguy:
@norwood this is great! Thanks.
Is the plan to eventually include checkstyle in this, too?
Ideally these checks should fail the build (warnings are too easy to ignore). I know this will be difficult to begin with, but would be great if that is the end goal.

pom.xml
xvrl: may be worth a comment in the pom.

xvrl:
@norwood it might be worthwhile to create a separate pom whose sole function is to act as parent pom for all our poms. This is mainly to avoid inadvertently adding something to common that could affect downstream repositories. Here's an example of what I did for Druid https://github.com/druid-io/oss-parent/blob/master/pom.xml

norwood:
@dguy yes. thats the end goal. 👶 steps

@xvrl yeah, the idea was to use the base pom in common as all of our cp repos depend on it (i think)

example: https://github.com/confluentinc/rest-utils/pull/60

pom.xml
```diff
@@ -113,4 +132,82 @@
...
+            <reporting>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>findbugs-maven-plugin</artifactId>
+                        <version>${findbugs-maven-plugin.version}</version>
+                        <configuration>
+                            <findbugsXmlOutput>true</findbugsXmlOutput>
+                            <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
+                            <xmlOutput>true</xmlOutput>
```
cmccabe: I like the HTML output, personally.  Unless you're going to parse it with a tool or something.

norwood: yeah, the idea is to have this be parsed here. https://jenkins.confluent.io/job/common-pr/36/

i can add html output too if youd prefer (and if thats possible)

pom.xml
```diff
@@ -113,4 +132,82 @@
...
+                            <report>scm</report>
+                        </reports>
+                    </reportSet>
+                </reportSets>
+            </plugin>
+        </plugins>
+    </reporting>
+    <profiles>
+        <profile>
+            <id>jenkins</id>
```
ewencp: Is everything under this profile too expensive to run for normal builds? I ask since I imagine we'd want to eventually turn some of these checks (e.g. jacoco coverage) into blockers for check-in, in which case having them as part of the normal build might be a bit nicer since it catches issues earlier.

utils/pom.xml
```diff
@@ -30,6 +30,7 @@
     <properties>
     </properties>
 
+    <version>3.2.1-SNAPSHOT</version>
```
norwood: even better we don't need this. it inherits the parent pom's version

utils/pom.xml
```diff
@@ -42,6 +43,11 @@
             <version>${slf4j.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.zookeeper</groupId>
+            <artifactId>zookeeper</artifactId>
+            <version>3.4.8</version>
```
norwood: moved this up to parents managed deps

pom.xml
norwood: i think so for now.

also have it tied to reporting right now. will need to get it bound to test which is another bit of change :/

norwood:
moved this to master branch

utils/pom.xml
```diff
@@ -42,19 +41,12 @@
             <version>${slf4j.version}</version>
```
xvrl: might want to remove the version number here to be consistent with the other modules

norwood:
retest this please

pom.xml
cmccabe: I would prefer to have html output as well, but I don't feel strongly about it.  If it's not convenient to get both, just do what's easier for the tooling.

pom.xml
```diff
@@ -113,4 +155,82 @@
             </plugin>
         </plugins>
     </build>
+    <reporting>
+        <!--https://maven.apache.org/plugins/maven-site-plugin/maven-3.html#Classic_configuration_Maven_2__3-->
```
xvrl: bumping maven site plugin to a more recent version should be enough

PR #44, opened 2017-03-06
CP-411 we need to allow keys that match synonym[0] through

norwood:
without this we will filter everything in a synonym group if all we get is something in synonym[0]

### DIFF
--- a/config/src/main/java/io/confluent/common/config/ConfigUtils.java
+++ b/config/src/main/java/io/confluent/common/config/ConfigUtils.java
@@ -49,7 +49,10 @@ public class ConfigUtils {
     HashSet<String> synonymSet = new HashSet<>();
     for (String[] synonymGroup: synonymGroups) {
       for (String synonym : synonymGroup) {
-        synonymSet.add(synonym);
+        if (!synonymSet.add(synonym)) {
+          // TODO: we should consider doing more formal validations around this
+          log.warn("reused synonym={}", synonym);
+        }
       }
     }
     // Properties is a very old Java class which uses Enumeration to iterate through keys rather
@@ -74,6 +77,9 @@ public class ConfigUtils {
       }
       if (deprecated.isEmpty()) {
         // No deprecated key(s) found.
+        if (props.containsKey(target)) {
+          newProps.put(target, props.get(target));
+        }
         continue;
       }
       String synonymString = deprecated.get(0);
--- a/config/src/test/java/io/confluent/common/config/ConfigUtilsTest.java
+++ b/config/src/test/java/io/confluent/common/config/ConfigUtilsTest.java
@@ -21,6 +21,8 @@ import org.junit.Test;
 import java.util.Properties;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


ijuma:
cc @cmccabe

config/src/main/java/io/confluent/common/config/ConfigUtils.java
```diff
@@ -48,8 +48,8 @@ public static Properties translateDeprecated(Properties props, String[][] synony
     // Properties object.
     HashSet<String> synonymSet = new HashSet<>();
     for (String[] synonymGroup: synonymGroups) {
-      for (String synonym : synonymGroup) {
-        synonymSet.add(synonym);
+      for (int i = 1; i < synonymGroup.length; i++) {
+        synonymSet.add(synonymGroup[i]);
```
cmccabe: Bypassing this here means that we won't eliminate cases where the different synonyms are set to different values.  I think it would be better to fix this bug this way:

>      if (deprecated.isEmpty()) {
>        // No deprecated key(s) found.
+       newProps.put(target, props.get(target));
>        continue;
>      }

config/src/test/java/io/confluent/common/config/ConfigUtilsTest.java
```diff
@@ -61,5 +63,20 @@ public void testTranslateDeprecated() throws Exception {
...
+
+  }
+
+  @Test
+  public void testAllowsNewKey() throws Exception {
+    Properties props = new Properties();
+    props.setProperty("foo.bar", "baz");
+    Properties newProps = ConfigUtils.translateDeprecated(props, new String[][]{
+        {"foo.bar", "foo.bar.deprecated"},
+        {"chicken", "rooster", "hen"},
```
cotedm: Do we need these extra farm animals? I don't think they are used in this test.

norwood: not necessary, but actually having them in there tests that they don't affect the other props

norwood:
@cmccabe can you take another look plz

config/src/main/java/io/confluent/common/config/ConfigUtils.java
```diff
@@ -49,7 +49,10 @@ public static Properties translateDeprecated(Properties props, String[][] synony
     HashSet<String> synonymSet = new HashSet<>();
     for (String[] synonymGroup: synonymGroups) {
       for (String synonym : synonymGroup) {
-        synonymSet.add(synonym);
+        if (!synonymSet.add(synonym)) {
+          // TODO: we should consider doing more formal validations around this
+          log.warn("reused synonym={}", synonym);
```
ewencp: This is one of only a couple of lines not covered by tests. Is there something small and simple we could do to exercise it?

norwood: yes. also added another test for the other branch lower down that was missed

norwood:
changed this to 3.2.x this needs to go out in a hotfix (assuming we make one)

PR #46, opened 2017-03-08
Fix handling of ConfigDef defaults and make NO_DEFAULT_VALUE safer as an Object instead of String.

ewencp:


ewencp:
Instead of reverting the change that introduced the bug, I cleaned up the behavior. The original patch was to address some findbugs warnings.

norwood:
shame

👍 🇮🇩

PR #48, opened 2017-03-22
Set the parent of the build-tools module so packaging can correctly pick up binary artifacts.

ewencp:


ewencp:
Ok, turns out this doesn't work since the parent pom needs the build-tools to execute but the parent always executes before children so we can't bootstrap. It passed locally only because I already had build-tools installed.

I guess I'll try to figure out how to get maven-assembly-plugin to behave with the extra module.

ewencp:
Turns out we were using moduleSets here but dependencySets everywhere else and moduleSets come with some caveats.

Merging as the current state of things will break all packaging builds on this branch, but review still welcome.

PR #49, opened 2017-03-27
moving even more deps from downstream to common

norwood:
this covers at least rest-utils/schema-registry/REDACTED

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,10 +49,20 @@
         <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
         <jacoco-maven-plugin.version>0.7.9</jacoco-maven-plugin.version>
         <junit.version>4.11</junit.version>
+        <kafka.version>0.11.0.0-SNAPSHOT</kafka.version>
+        <kafka.scala.version>2.11</kafka.scala.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -101,6 +112,73 @@
...
+                <version>${kafka.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.kafka</groupId>
+                <artifactId>connect-json</artifactId>
+                <version>${kafka.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.kafka</groupId>
+                <artifactId>connect-file</artifactId>
```
ewencp: Does anything actually depend on this? It's a connector, so I can't imagine it has much useful in it.

norwood: https://github.com/confluentinc/schema-registry/blob/2d5a9ad74187e1465242207330e35b038da429e4/package-kafka-serde-tools/pom.xml#L43-L63

maybe there is a better way to do this? i just went on autopilot doing this tbh

ewencp: Ah, I didn't remember about that. Clearly I was trying to be nicer to myself as a developer back then, these days I just assume you have to deal with classpath stuff. We could clean that up, but fine to leave this as is for now.

PR #51, opened 2017-04-03
Integration test pom fixins

norwood:
this does a couple things
1. renames current common parent/module pom to `common-parent`
2. creates a replacement parent with the old name for downstream deps to use (this is the `parent` module)
3. re-add the integration test configs to `common-parent`

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -1,5 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <!--~
-  ~ Copyright 2015 Confluent Inc.
+  ~ Copyright 2017 Confluent Inc.
   ~
   ~ Licensed under the Apache License, Version 2.0 (the "License");
   ~ you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


parent/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+        <version>3.3.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>common</artifactId>
+    <packaging>pom</packaging>
+    <name>Common</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.confluent</groupId>
```
ewencp: Document why this is included here? And maybe also why we have this pom at all?

parent/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+
+    <parent>
+        <artifactId>common-parent</artifactId>
+        <groupId>io.confluent</groupId>
+        <version>3.3.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>common</artifactId>
+    <packaging>pom</packaging>
+    <name>Common</name>
```
ewencp: Naming is a bit weird here. We've got artifactId common with name Common and the top level pom here is artifactId common-parent with name common (which makes maven output showing the reactor build ordering confusing).

PR #55, opened 2017-05-12
Remove ClassFanOutComplexity check

theduderog:
As discussed in the `#code-quality` slack channel, this check does more harm than good.

I think this removes the ClassFanOutComplexity check.


dguy:
LGTM

PR #62, opened 2017-06-08
use common jenkins configs

norwood:


Jenkinsfile
```diff
@@ -1,35 +1,5 @@
...
-        break;
-      case 'UNSTABLE':
-        slackSend(channel: '#clients-eng', color: 'YELLOW', message: "${env.JOB_NAME} - #[${env.BUILD_NUMBER}] Unstable <${env.BUILD_URL}|(Open)>", teamDomain: 'confluent')
-        break;
-      case 'FAILURE':
-        slackSend(channel: '#clients-eng', color: 'bad', message: "${env.JOB_NAME} - #[${env.BUILD_NUMBER}] Failure <${env.BUILD_URL}|(Open)>", teamDomain: 'confluent')
-        break;
-    }
-  }
+common {
```
maxzheng: Can u provide info on where 'common' is defined? I am not familiar with this feature u are trying to use...

norwood:
sorry. its from https://github.com/confluentinc/jenkins-common/blob/master/vars/common.groovy 

in our jenkins org config we can load that as a lib. this way each repo just needs to have overrides for things, but common steps are all consolidated in one place

PR #63, opened 2017-07-19
Use --release instead of --source/--target if compiling with Java 9

ijuma:
This allows one to use Java 9 and compile for older releases safely. See: http://openjdk.java.net/jeps/247

`mvn compile` and `mvn test` succeed when running with Java 9, but I saw some exceptions when running `mvn` with debug enabled. We'll probably have to bump the version of a number of plugins for everything to work correctly with Java 9. We can do that in a separate PR though.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
         <kafka.version>0.11.1.0-SNAPSHOT</kafka.version>
         <kafka.scala.version>2.11</kafka.scala.version>
         <maven-assembly.version>2.6</maven-assembly.version>
-        <maven-compiler-plugin.version>2.5.1</maven-compiler-plugin.version>
+        <maven-compiler-plugin.version>3.6.1</maven-compiler-plugin.version>
         <maven-deploy.version>2.8.2</maven-deploy.version>
         <maven-enforcer-plugin.version>1.4.1</maven-enforcer-plugin.version>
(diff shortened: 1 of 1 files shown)


ijuma:
@xvrl will probably be most interested in this one, but also cc'ing @norwood and @ewencp so that they're aware.

ijuma:
@xvrl, agreed. Not sure how the JDK is installed in Jenkins, it may be that Java 9 is not yet available in the repo used for other JDKs.

ewencp:
@ijuma @xvrl We'd need to update the docker images, pretty much everything is landing in docker. We don't even have jdk8 builds atm either, we just use the jdk7 baseline. We have jdk8 images, but really only have builds running against jdk7.

Updating the images is probably straightforward, but we'd want to decide now if we want one big images w/ all the jdks and require jobs to select between them or multiple images w/ individual jdks such that specifying the image (via jenkins label) gives you what you want. Since nearly all of the Java jobs are now on a shared baseline Jenkinsfile, I think making the change for that would probably just work for all the Java jobs.

ijuma:
@ewencp Thanks for the explanation. I guess the answer depends on how much of a benefit we get from keeping images small and the additional overhead of managing 3 images (if any).

xvrl:
@ewencp yes, I didn't mean this needed to happen right away. I was just saying that until we have other JDKs in our build pipeline we aren't really exercising this new build profile.

PR #64, opened 2017-08-22
add build tools to the packaging for common

norwood:


package/pom.xml
```diff
@@ -34,6 +34,11 @@
     <dependencies>
         <dependency>
             <groupId>io.confluent</groupId>
+            <artifactId>build-tools</artifactId>
+            <version>3.3.1-SNAPSHOT</version>
```
aayars: Is version required? Just asking cuz this is one more thing to bump.

norwood: common depends on build-tools, so we end up with circular refs if i try to do anything smarter here :/

PR #65, opened 2017-08-23
bump to align with version on this branch

norwood:


norwood:
going to  merge this since it is breaking builds

PR #66, opened 2017-09-14
Add upstream project so build are triggered automatically

ewencp:


ewencp:
I can set the value correctly for master when I merge up.

PR #68, opened 2017-09-16
Disable the security manager during tests.

ewencp:


ewencp:
@norwood I noticed we're not getting coverage information from this build despite being on the common pom. Turns out it is because of this override https://github.com/confluentinc/kafka-connect-elasticsearch/blob/master/pom.xml#L151-L157 which doesn't match the settings in the common pom.

This seems like the easiest way to avoid this issue for all projects, but I'm not sure if this is really the best place for it, but I can't think of a reason it would be a problem. I think this is happening for at least a few other repositories as well, e.g. I see `Number of found exec files for pattern **/**.exec: 0` in the security plugins as well (and a report showing 0 line coverage despite there being tests).

norwood:
can the downstream users of this not just include `{argLine}` in the their configuration ?

ewencp:
Yeah. I was looking at the number of times `maven-surefire-plugin` appears in downstream jobs and trying to figure out what stuff we might just want to bubble up here to avoid having to override (this particular config didn't seem like any downstream projects would care about having toggled this way by default).

Now that I've gone through more of them, it looks like only a few override this config and it looks like some of the references aren't even useful, they are just re-including the plugin. I'll close out this and have filed the [corresponding fix to the ES repo](https://github.com/confluentinc/kafka-connect-elasticsearch/pull/122).

PR #69, opened 2017-09-19
Add a test utility for creating temporary directories

ewencp:


utils/src/main/java/io/confluent/common/utils/Utils.java
```diff
@@ -356,4 +363,34 @@ public static Properties loadProps(String filename) throws IOException {
...
+   * @param file The root file at which to begin deleting
+   */
+  public static void delete(final File file) throws IOException {
+    if (file == null)
+      return;
+    Files.walkFileTree(file.toPath(), new SimpleFileVisitor<Path>() {
+      @Override
+      public FileVisitResult visitFileFailed(Path path, IOException exc) throws IOException {
+        // If the root path did not exist, ignore the error; otherwise throw it.
+        if (exc instanceof NoSuchFileException && path.toFile().equals(file))
```
norwood: you need `{}` on this if statement

ewencp: We have a bunch of stuff copied from AK that we have ignored checkstyle on. This is sort of that, though I didn't copy all the variants/the complete flexibility of the part I did copy.

I've adjusted to match style, but we should probably either decide to abandon the AK compatibility or avoid making more formatting changes in the future.

utils/src/main/java/io/confluent/common/utils/TestUtils.java
```diff
@@ -0,0 +1,54 @@
...
+public class TestUtils {
+
+  /**
+   * Create a temporary directory. The directory and any contents will be deleted when the test
+   * process terminates.
+   */
+  public static File tempDirectory() {
+    final File file;
+    try {
+      file = Files.createTempDirectory("confluent").toFile();
```
trejkaz: Is there a guarantee that a temp directory will always be on the default filesystem?

ewencp: @trejkaz What do you mean by "default" filesystem? The same as `/` or are you thinking in terms of a non-posix platform?

Short version is no, there shouldn't be any such assumption here. For example, `/tmp` could easily be on a standalone, memory-backed disk.

trejkaz: "default" in the sense of `FileSystems.getDefault()`, and the note on `Path.toFile()` saying that it will throw `UnsupportedOperationException` if it isn't on the default one.

But I just re-read the docs on `createTempDirectory`, and indeed, if you don't use the overload taking `Path`, the file it creates _is_ created on the default filesystem. Which is good, this is another assumption that can be used to permit some calls to toFile(). (I'm in the middle of a several year war to stomp them out in our own project.)

PR #72, opened 2017-09-27
CP-497 fix not being able to run unit tests in IDE

xvrl:


norwood:
you can also set this up in your ide: https://stackoverflow.com/questions/24115142/intellij-error-when-running-unit-test-could-not-find-or-load-main-class-suref

@ewencp wasn't there something else that just changed and was `argLine` related?

xvrl:
agree there are workarounds, but unless we really need to override the command line, I figured it would be simpler to use system properties? I didn't find any other references to argLine in common at least.

ewencp:
there were changes in other repositories when i was converting things to common pom because they had some other settings as well. this might break them, i don't really know anything about the argline stuff and i can't remember which repos they were -- maybe kafka-connect-elasticsearch, but there may have been others as well

xvrl:
hmm, looks like surefire can't set those arguments before the vm starts anyway so I'm going to close this

ewencp:
for reference, i've just dealt with the horrible route of toggling the checkbox in intellij until we can get everything in CP using the same approach... and i agree it is quite annoying

PR #73, opened 2017-09-29
Enable unused import checkstyle check

ewencp:


ewencp:
I put this on 3.3.x, but we could target to something later if people don't feel comfortable adding it to a bugfix branch.

Fallout may not be as bad as I thought, took running this against common, rest-utils, schema-registry, and kafka-rest before I finally found a violation. I'll enable it when I can watch jenkins for the fallout and file follow up PRs across all affected branches to get us into a good state.

PR #74, opened 2017-10-07
Use maven-failsafe-plugin for integration-tests

xvrl:
Moving integration tests to failsafe plugin will accomplish several things:
- Shorter development iteration, since integration-test phase comes after package phase
- Unit tests will be run first before any of the integration tests
- A single failing integration test will not prevent other integration tests from running
- Separate test reports for integration and unit tests

This may impact any downstream projects that depend on shaded jars in their integration tests. If you have one of those please come talk to me.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -327,6 +327,11 @@
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-failsafe-plugin</artifactId>
+                    <version>${maven-surefire-plugin.version}</version>
+                </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-jar-plugin</artifactId>
                     <version>${maven-jar-plugin.version}</version>
                 </plugin>
@@ -352,6 +357,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <argLine>@{argLine} -Djava.awt.headless=true</argLine>
+                    <excludedGroups>io.confluent.common.utils.IntegrationTest</excludedGroups>
                 </configuration>
                 <executions>
                     <execution>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -331,36 +355,51 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <configuration>
+                    <groups>io.confluent.common.utils.IntegrationTest</groups>
+                    <!-- by default failsafe only includes class names that start or end in IT,
+                         we use regular test class names with @Category and the group above -->
+                    <includes>
+                        <include>**/Test*.java</include>
```
ewencp: Is there any reason not to just use `**/*.java`? The naming restriction seems like it's just a recipe for accidentally not running a test and we already filter by the annotation.

xvrl: I just copied the defaults used by `maven-surefire-plugin` to keep things consistent.

ewencp: Oh, that makes sense. Now I am wondering if we have any unit tests that are not being executed... I sort of assumed anything with the `@Test` annotation would be executed, guess I need to actually learn the java ecosystem's tools properly at some point.

xvrl:
Arguably I should probably add back the failsafe defaults as well, just to
make sure someone isn’t caught offguard in the future

pom.xml
```diff
@@ -331,36 +355,57 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <configuration>
+                    <argLine>@{argLine} -Djava.awt.headless=true</argLine>
+                    <groups>io.confluent.common.utils.IntegrationTest</groups>
+                    <!-- by default failsafe only includes class names that start or end in IT,
+                         we use regular test class names with @Category and the group above -->
+                    <includes>
```
norwood: do we actually need anything here? if so, maybe just `**/*.java`?

xvrl: as I mentioned to @ewencp here https://github.com/confluentinc/common/pull/74#discussion_r143609458, the surefire plugin defaults to those same classes, so I think it makes sense to stick to the default unit test class names

PR #76, opened 2017-10-16
MINOR: Add reset to MockTime class.

kkonstantine:


kkonstantine:
When debugging and writing specific tests, the ability to reset the mocked time class to a specific point in time (e.g. Epoch, present) seems reasonable. 

This is targeting specifically the mock and is not becoming part of the API, allowing only the code that needs it to have to adapt.

kkonstantine:
It's useful when debugging and you need to start from an easy starting point (epoch) instead of always subtracting the starting time. 

Also this starting time is not flexible enough be being set only at the creation time of the mock object (there are cases where the mock is static and not created per test case). 

I agree that this will not be used in test cases that use a mock per test run. But for the above  cases (need to debug and need to use a static mock across tests) this might be useful. Small fraction of cases but annoying to not be able to.

kkonstantine:
ok to test

kkonstantine:
Cool, thanks. Yeah, limited usability but not to bad to add either.

PR #77, opened 2017-10-17
Checkstyle overrides are needed in reporting config also

hachikuji:
If you build site with the jenkins profile, you can see in `site/checkstyle-aggregate.html` that it uses `sun_checks.xml` instead of our own checkstyle config. Not sure if there are better solutions, but overriding the config in the profile seems to work. One crappy thing is that we may need to override the suppressions location for each project separately... Not sure if we can find a way to add `checkstyle/suppressions.xml` to the list of suppressions here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -506,6 +506,23 @@
                             </execution>
                         </executions>
                     </plugin>
+                    <plugin>
+                      <groupId>org.apache.maven.plugins</groupId>
+                      <artifactId>maven-checkstyle-plugin</artifactId>
+                      <version>2.17</version>
+                      <dependencies>
+                        <dependency>
+                          <groupId>io.confluent</groupId>
+                          <artifactId>build-tools</artifactId>
+                          <version>${confluent.version}</version>
+                        </dependency>
+                        <dependency>
+                          <groupId>com.puppycrawl.tools</groupId>
+                          <artifactId>checkstyle</artifactId>
+                          <version>6.19</version>
+                        </dependency>
+                      </dependencies>
+                    </plugin>
                 </plugins>
             </build>
             <reporting>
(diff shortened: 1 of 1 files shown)


hachikuji:
@norwood @ewencp Either of you care to take a look?

hachikuji:
One solution for the problem with checkstyle suppressions is to add a property for the file location. I tried this out locally and it seems to work.

pom.xml
```diff
@@ -67,6 +67,8 @@
         <slf4j.version>1.7.25</slf4j.version>
         <zkclient.version>0.10</zkclient.version>
         <zookeeper.version>3.4.10</zookeeper.version>
+        <checkstyle.config.location>checkstyle/checkstyle.xml</checkstyle.config.location>
+        <checkstyle.suppressions.location>checkstyle/common-suppressions.xml</checkstyle.suppressions.location>
```
ewencp: in this case i think we want to separate these out. we actually already have these named differently in children -- they are named just `checkstyle/suppressions.xml` in children projects. e.g. https://github.com/confluentinc/kafka-connect-hdfs/blob/master/checkstyle/suppressions.xml I think we've already been overriding them manually without much complaint, e.g. https://github.com/confluentinc/kafka-connect-hdfs/blob/master/checkstyle/suppressions.xml but including a property for it wouldn't hurt.

maxzheng:
Can we backport this to 3.3.0-post / 3.3.x as Jenkinsfile was used since then? That will be needed when I start to build post branches without the `deploy` step for all repos. I suppose it should go into 3.3.0-post branch and then pint merged up to all other branches.

pom.xml
hachikuji: Right, but overriding is a bit more difficult now since you also have to do it for the jenkins profile.

hachikuji:
@maxzheng Ack. I will close this and retarget it for 3.3.x.

PR #78, opened 2017-10-18
Checkstyle overrides are needed in reporting config also

hachikuji:


hachikuji:
cc @maxzheng @ewencp

norwood:
what exactly is this solving?

hachikuji:
@norwood See the description in #77. You need the config overrides for the reports in the jenkins profile to work correctly. If not, the jenkins build always fails as "unstable."

hachikuji:
For example, see the build here: https://jenkins.confluent.io/job/confluentinc-pr/job/kafka-broker-plugins/job/PR-9/2/console. It fails because it is applying the wrong checkstyle rules:
```
[INFO] Generating "Checkstyle" report    --- maven-checkstyle-plugin:2.17
[INFO] There are 850 errors reported by Checkstyle 6.19 with sun_checks.xml ruleset.
```

norwood:
can we just not run `site` in the `mvn` command in `common.groovy`?

maxzheng:
@norwood site is running the checkstyle if it is setup for the repo, and other reports, like dependency. Not sure about the other reports, but we do need checkstyle. If there is a way to run checkstyle (without site) only if the project is setup to do so, then we can do that too -- but I don't know if that's possible with mvn. Maybe run validate instead? That is an alternative when deploy is skipped (either run deploy or validate). @ewencp  @norwood WDYT? In any case, this is still an issue for site that should be fixed too.

norwood:
checkstyle is setup to run @validate https://github.com/confluentinc/common/blob/master/pom.xml#L423

jacoco runs @verify https://github.com/confluentinc/common/blob/master/pom.xml#L502

so maybe we could just run `verify` instead of `site`?

my main concern is that all our downstream builds break cause of this. i'd prefer if we can not require everyone to update

maxzheng:
@norwood @ewencp Replacing `site` with `verify` works me as long as you don't need the other reports. I will put up a separate PR for that. As for this PR, @hachikuji , can u confirm that this won't break downstream projects? if it doesn't break / doesn't require any changes in downstream projects (optional is ok, but not required), then it is still worth fixing and checking in for future. As it is no longer a concern for me, I leave it up to you guys. :)

maxzheng:
@norwood Unfortunately, findbug needs the report: 

```[FINDBUGS] Collecting findbugs analysis files...
[FINDBUGS] Searching for all files in /home/jenkins/workspace/ntinc_jenkins-common_master-BWNTZ3YGHPVE7NDR54DIQLTAXP6HKJB7F7QLW7WI2U2JRXZDJOLA/tests/java that match the pattern **/findbugsXml.xml
[FINDBUGS] No files found. Configuration error?```

It still shows the warnings but there is no fancy report: ```FindBugs: 0 warnings.
9 fixed warnings
During parsing an error has been reported.``` https://jenkins.confluent.io/job/confluentinc/job/jenkins-common/job/master/9/

norwood:
k. lets merge this then

hachikuji:
I can't see why it would break downstream projects. In the worst case, the jenkins profile won't pick up suppressions which have been overridden. That said, if it does break anything, we can just revert it.

hachikuji:
Yay! My PR build is working now: https://jenkins.confluent.io/job/confluentinc-pr/job/kafka-broker-plugins/job/PR-10/6/.

maxzheng:
@hachikuji I cherry-picked it to 3.3.0-post, so if we need to make any changes, let's make it there and then pint merge. Thanks Jason for figuring this out and fixing it! 💯

maxzheng:
Actually, can't work for 3.3.0-post unless we re-build/deploy common, which I am not sure if that's safe to do for something that is released. So checked in fix on jenkins-common side as well.

PR #79, opened 2017-10-19
Update maven plugin versions and powermock

ijuma:
The main motivation is Java 9 support, but also fixed the version
for some commonly used plugins that were previously dependent
on the Maven version being used.

Also fix some inconsistencies: we now use pluginManagement
backed by properties for all plugin versions.

### DIFF
--- a/package/pom.xml
+++ b/package/pom.xml
@@ -55,7 +55,7 @@
         <plugins>
             <plugin>
                 <artifactId>maven-assembly-plugin</artifactId>
-                <version>2.5.2</version>
+                <version>3.0.0</version>
                 <executions>
                     <execution>
                         <id>package-assembly</id>
--- a/package/pom.xml
+++ b/package/pom.xml
@@ -35,7 +35,7 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>build-tools</artifactId>
-            <version>4.1.0-SNAPSHOT</version>
+            <version>${confluent.version}</version>
         </dependency>
         <dependency>
             <groupId>io.confluent</groupId>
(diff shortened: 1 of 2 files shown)


package/pom.xml
```diff
@@ -55,7 +55,7 @@
         <plugins>
             <plugin>
                 <artifactId>maven-assembly-plugin</artifactId>
-                <version>2.5.2</version>
+                <version>3.0.0</version>
```
ewencp: this should probably be using the `maven-assembly.version` property

pom.xml
```diff
@@ -312,8 +313,8 @@
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-deploy-plugin</artifactId>
-                    <version>${maven-deploy.version}</version>
+                    <artifactId>maven-clean-plugin</artifactId>
+                    <version>3.0.0</version>
```
ewencp: all of these plugin changes should also be using `maven-<whatever>.version` variables.

package/pom.xml
ijuma: It seemed like properties were used a bit inconsistently. Seems like you are suggesting that we should always use properties? If so, is the motivation to make them easier to update?

ewencp: yeah, they probably are used inconsistently. there are a couple of reasons. first, we ended up in a few places where versions were getting inconsistent. obviously some are not repeated, but others are and it makes updating easier. the other reason is that properties get inherited from parent poms, so defining them is important, but then sticking to only using properties ensures we're consistent everywhere.

ijuma:
@ewencp, I believe I've addressed your comments?

ijuma:
Haha, I did indeed only update the root pom.xml. My bad.

ijuma:
@ewencp can you please take another look?

PR #80, opened 2017-10-26
Fix powermock version.

ewencp:


ewencp:
Not really sure, I missed it too :) Only way I can think of is to chain all downstream builds (or only introduce dependencies where they are first used).

PR #87, opened 2018-01-03
ST-654: Fail builds on High findbugs warnings

ewencp:


ewencp:
I set the threshold to high initially so we can fix the most critical things first and see whether a lower threshold is useful later. I believe we'll still get the full reports since the `reporting` section is distinct, but we'll see in the PRs here. I still need to go through and fix up all the High warnings in all the projects before merging this.

ewencp:
All our significant dev is on master now anyway and since I need to fix things in other repos, I was only turning it on for master so I don't need to make those changes on 4.0.x. So far none of them are that significant, but I am seeing some like relying on default charsets that make me a bit more nervous about changing in a bugfix branch.

ewencp:
Also, I've now added the annotations into dependencyManagement since I need to suppress some cases in downstream projects.

ewencp:
And now I'm back to the original with the dependency addition moved to #89 so #89 can be merged first, unblock downstream fixes that need the dependency, and then finally unblock this once all the downstream fixes are made.

xvrl:
@ewencp since this got merged the blueway builds now have up to 27 findbugs reports instead of just one

pom.xml
```diff
@@ -564,10 +563,26 @@
                         <dependency>
                           <groupId>com.puppycrawl.tools</groupId>
                           <artifactId>checkstyle</artifactId>
-                          <version>6.19</version>
+                          <version>${checkstyle.version}</version>
```
xvrl: version should be defined once and for all under pluginManagement so it doesn't have to be specified in plugin at all.

pom.xml
```diff
@@ -564,10 +563,26 @@
...
                         </dependency>
                       </dependencies>
                     </plugin>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>findbugs-maven-plugin</artifactId>
```
xvrl: I believe findbugs needs to be configured like checkstyle to depend on build-tools so it's only run once per build in a multi-module setup. According to https://gleclaire.github.io/findbugs-maven-plugin/examples/multi-module-config.html, it should be just like checkstyle, but we that article uses an extension instead of it a plugin dependency, so pick whichever works.

pom.xml
ewencp: follow up in https://github.com/confluentinc/common/pull/92

pom.xml
ewencp: this turns out to be a different issue, but follow up in https://github.com/confluentinc/common/pull/92

PR #88, opened 2018-01-04
Map type config to support metrics tags

mageshn:


config/src/main/java/io/confluent/common/config/ConfigDef.java
```diff
@@ -336,6 +336,27 @@ else if (value instanceof String)
           } else {
             throw new ConfigException(name, value, "Expected a comma separated list.");
           }
+        case MAP:
+          if (value instanceof Map) {
+            return value;
+          } else if (value instanceof String) {
+            if(trimmed.isEmpty()) {
```
edenhill: inconsistent spacing, should be `if (`

config/src/main/java/io/confluent/common/config/ConfigDef.java
```diff
@@ -336,6 +336,27 @@ else if (value instanceof String)
...
+              return Collections.emptyMap();
+            } else {
+              Map<String, String> map = new HashMap<>();
+              List<String> entries = Arrays.asList(trimmed.split("\\s*,\\s*", -1));
+              for ( String entry : entries) {
```
edenhill: extra space inserted before `String`

config/src/main/java/io/confluent/common/config/ConfigDef.java
```diff
@@ -336,6 +336,27 @@ else if (value instanceof String)
...
+                String[] keyValue = entry.split("\\s*:\\s*", -1);
```
edenhill: same comment as in the other PR: "=" as delimiter seems more natural. (but I don't really have anything to back that up)

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
```diff
@@ -77,6 +79,7 @@ public void testBasicTypes() {
     props.put("h", "FalSE");
     props.put("i", "TRUE");
     props.put("j", "password");
+    props.put("k", "k1:v1,k2:v2");
```
edenhill: add a test for single key-value (no , separator)

config/src/main/java/io/confluent/common/config/ConfigDef.java
```diff
@@ -336,6 +336,27 @@ else if (value instanceof String)
...
+                if (keyValue.length != 2) {
+                  throw new ConfigException("Map entry should be of form <key>:<value");
```
wicknicks: missing a ">" after value.

mageshn:
@ewencp this is to provide ability to configure metrics tags. This will help add additional tags for monitoring systems. Kafka has a similar config listener.security.protocol.map but then its defined as a String instead of a Map.

config/src/main/java/io/confluent/common/config/ConfigDef.java
mageshn: I based it off the config in Kafka listener.security.protocol.map. The separator here is ':'

mageshn:
@edenhill ping :)

PR #89, opened 2018-01-04
Add findbugs annotations to dependencyManagement for downstream projects

ewencp:


ewencp:
I was going to do this in #87, but after making PRs for other repos I realize it will be better to handle this separately first so we can get passing builds on those PRs, get them merged, then finally merge #87 once I've gotten all downstream failures fixed.

PR #91, opened 2018-01-06
Consolidating common-docker into common

mohnishbasha:
Consolidated: common-docker:master -> common:master

- assembly-plugin-boilerplate
- utility-belt
- cp-base

How did I test:
- mvn clean  -Dmaven.test.skip=true
- was able to build locally.


base/setup.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
+
+from pip.req import parse_requirements
+import setuptools
+
+remote_requirements = '\n'.join(str(r.req) for r in parse_requirements("requirements.txt", session='dummy') if r.req)
+
+setuptools.setup(
+    name='common-tests',
+    version='4.0.0',
```
aayars: Might as well use 4.1.0.

Jenkinsfile
```diff
@@ -6,3 +6,11 @@ common {
   // change across version branches.
   upstreamProjects = 'kafka-trunk'
 }
+
+docker_oraclejdk8 {
+    dockerRegistry = '368821881613.dkr.ecr.us-west-2.amazonaws.com/'
+    dockerRepos = ['confluentinc/cp-base']
+    slackChannel = '#tools-eng'
+    upstreamProjects = ['confluentinc/confluent-docker-utils']
+    withPush = true
```
aayars: - change withPush to dockerPush (withPush was deprecated)
- use 'clients-eng' for docker_oraclejdk8's slackChannel
- add 'kafka-trunk' to docker_oraclejdk8's upstreamProjects, preserving the old comment from common{}.
- add nodeLabel = 'docker-oraclejdk7' to the docker_oraclejdk8 block
- remove the common{} block

pom.xml
```diff
@@ -80,6 +80,20 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <GIT_COMMIT>unknown</GIT_COMMIT>
```
aayars: Add a comment above this line to break things up, like `<!-- docker build support -->`

src/assembly/package.xml
```diff
@@ -0,0 +1,52 @@
...
+        <format>dir</format>
+    </formats>
+    <includeBaseDirectory>false</includeBaseDirectory>
+    <fileSets>
+        <fileSet>
+            <directory>${project.parent.basedir}</directory>
+            <outputDirectory>share/doc/${project.artifactId}/</outputDirectory>
+            <includes>
+                <include>version.txt</include>
+                <include>COPYRIGHT*</include>
```
ewencp: does this cover all the general stuff? off the top of my head, I would have thought at least `LICENSE` would be included here as well.

src/assembly/package.xml
```diff
@@ -0,0 +1,52 @@
...
+        <fileSet>
+            <directory>${project.parent.basedir}</directory>
+            <outputDirectory></outputDirectory>
+            <includes>
+                <include>bin/*</include>
+            </includes>
+        </fileSet>
+        <fileSet>
+            <directory>${project.parent.basedir}/config</directory>
+            <outputDirectory>etc/${project.artifactId}</outputDirectory>
```
ewencp: pretty sure this is going to break down in at least a few cases. are we trying to completely generalize or just cover most cases? for example, blueway has at least 2 different packaging setups.

assembly-plugin-boilerplate/README.md
```diff
@@ -0,0 +1,10 @@
+# Assembly Plugin Docker Boilerplate
+
+This project uses the assembly plugin rules from resource.xml to package
+an artifact containing the raw contents of package.xml. See src/assembly,
```
ewencp: historically we have `package.xml`, `development.xml`, and `standalone.xml`. the standalone one was mainly for testing convenience as it introduces redistribution issues. development includes enough to test in a dev environment.

i mention this because the merging of docker stuff makes it unclear how this is intended to interact with traditional packaging and the dev cycle. how would this affect those other targets for downstream? is packaging going to be affected in anyway by including this in common?

pom.xml
```diff
@@ -80,6 +80,21 @@
...
+        <docker.skip-build>true</docker.skip-build>
+        <docker.skip-test>true</docker.skip-test>
+        <docker.registry>placeholder/</docker.registry>  <!-- Registry for this build. Must specify a value, or exec plugin throws NPE :((( -->
+        <docker.tag>${project.version}</docker.tag>  <!-- Tag for this build -->
+        <docker.upstream-registry>${docker.registry}</docker.upstream-registry>  <!-- Registry for base images -->
+        <docker.upstream-tag>${docker.tag}</docker.upstream-tag>  <!-- Tag for base images -->
+        <docker.test-registry>${docker.upstream-registry}</docker.test-registry>  <!-- Registry for integration test dependencies -->
+        <docker.test-tag>${docker.upstream-tag}</docker.test-tag>  <!-- Tag for integration test dependencies -->
+        <dockerfile-maven-plugin.version>1.3.6</dockerfile-maven-plugin.version>
+        <exec-maven-plugin.version>1.5.0</exec-maven-plugin.version>
```
ewencp: if we highlight `docker build support` these following 3 things seem like they might not fit that category? or at least even if they are being pulled in due to it, they would not be specific to it?

pom.xml
```diff
@@ -646,5 +686,151 @@
...
+                        <version>${exec-maven-plugin.version}</version>
+                        <executions>
+                            <execution>
+                                <id>create-licenses-for-docker</id>
+                                <phase>package</phase>
+                                <goals>
+                                    <goal>java</goal>
+                                </goals>
+                                <configuration>
+                                    <mainClass>io.confluent.licenses.LicenseFinder</mainClass>
```
ewencp: probably a separate discussion from this pr, but we should probably discuss how licenses are managed and how that fits into standard mgmt of our repos

Jenkinsfile
```diff
@@ -1,8 +1,12 @@
...
-  upstreamProjects = 'kafka-trunk'
+docker_oraclejdk8 {
+    // This is not strictly required in this repo, but is easiest here so we only maintain this Kafka upstream trigger in one
+    // place to trigger the rest of the pipeline. Note that we need to maintain the references to the jobs carefully since they
+    // change across version branches.
+    upstreamProjects = ['kafka-trunk', 'confluentinc/confluent-docker-utils']
+    dockerRegistry = '368821881613.dkr.ecr.us-west-2.amazonaws.com/'
+    dockerRepos = ['confluentinc/cp-base']
+    slackChannel = '#clients-eng'
+    nodeLabel = 'docker-openjdk7'
```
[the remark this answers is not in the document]
mohnishbasha: @aayars @ewencp 
https://github.com/confluentinc/jenkins-common/blob/master/vars/docker_oraclejdk8.groovy
I guess we gotta consolidate eventually, but docker_oraclejdk8 refers to the groovy lib in jenkins-common. 

@aayars , should this be nodeLabel: 'docker-oraclejdk8' instead, looks like we are overriding the nodeLabel in common?

assembly-plugin-boilerplate/README.md
mohnishbasha: @aayars @ewencp - We believe it wont affect, however this needs to be validated once the merge goes in. I spoke to Alex about this, in case we run into issues, we intend to revert it back.

assembly-plugin-boilerplate/pom.xml
```diff
@@ -0,0 +1,58 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+ xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+ http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <!--
```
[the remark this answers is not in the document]
mohnishbasha: sure, will remove this, we don't need this to be built using common-parent. Instead we have made common to depend on assembly. Building with parent was causing assembly issues. I will have this cleaned up.

pom.xml
mohnishbasha: sure.

pom.xml
mohnishbasha: @aayars @ewencp - I mostly moved this from common-docker over to common.

src/assembly/package.xml
```diff
@@ -0,0 +1,52 @@
...
+        <dependencySet>
+            <outputDirectory>share/java/${project.artifactId}</outputDirectory>
+            <useProjectArtifact>true</useProjectArtifact>
+            <!-- Warning: Enabling useTransitiveFiltering is potentially unsafe,
+                 cf. https://github.com/miguno/maven-assembly-transitive-deps-woes
+                 and https://issues.apache.org/jira/browse/MASSEMBLY-504.  If
+                 dependencies happen to be missing in the assembly you may want to
+                 disable transitive filtering. -->
+            <useTransitiveFiltering>true</useTransitiveFiltering>
+            <!-- <excludes>
```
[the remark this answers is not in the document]
mohnishbasha: @ewencp - I mostly brought these over from common-docker. May be we call should do a zoom.

Jenkinsfile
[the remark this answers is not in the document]
aayars: This needs to build on a Java 7 node for now.

As a separate change, that library is being renamed to just 'docker'. After that goes in, we can update all the downstream projects (I wouldn't block on it here.)

assembly-plugin-boilerplate/README.md
aayars: This is a good point, and probably worth nailing down as part of this PR. `rest-utils` brings in the assembly configs mentioned by @ewencp, and there is a fair amount of overlap between the "docker" assemblies and the "packaging" assemblies. There's also overlap with the license generation stuff between `common-docker` and `rest-utils`.

For projects which inherit from `rest-utils` I am using rest-utils' assemblies (used by packaging) instead of common-docker's... just because they were already there, and happen to be perfectly suitable for copying into an image.

I propose consolidating the assembly configs from `rest-utils` into `common`. We can update the Dockerfile in downstream projects to use that style of assembly (including license file generation stuff), instead of the `common-docker` style of assembly. It's possible that there's no change at all required for some projects.

pom.xml
aayars: I'd move exec-maven-plugin.version, licenses.version, and slf4j-api.version up to the first group of properties (in alphabetical order) since they're not specific to Docker.

Jenkinsfile
aayars: The nodeLabel should actually be 'docker-oraclejdk7'

mohnishbasha:
Staged: https://github.com/confluentinc/rest-utils/pull/83
Plan is to merge this changes.
- Address any downstream issues that might arise.
- Later consolidate assembly from rest-utils.

PR #92, opened 2018-01-09
Remove unnecessary dependencies from checkstyle plugin.

ewencp:


ewencp:
This originally started as a PR to fix how findbugs reports were working and just addressed one other minor issue @xvrl mentioned in https://github.com/confluentinc/common/pull/87/files. However, I found that issue needs to be fixed in a different repo https://github.com/confluentinc/jenkins-common/pull/35 so now this is just for the checkstyle issue.

pom.xml
```diff
@@ -319,7 +324,6 @@
                         <dependency>
                             <groupId>com.puppycrawl.tools</groupId>
                             <artifactId>checkstyle</artifactId>
-                            <version>${checkstyle.version}</version>
```
xvrl: this one doesn't necessarily need to be moved, it's part of pluginManagement already. My comment in #87 was referring to the lines https://github.com/confluentinc/common/blob/master/pom.xml#L563-L574. Since we already specify the plugin dependencies and their versions here (in pluginManagement), we should be able to remove the lines I linked to entirely.

pom.xml
```diff
@@ -569,7 +569,6 @@
                         <dependency>
                           <groupId>com.puppycrawl.tools</groupId>
                           <artifactId>checkstyle</artifactId>
-                          <version>${checkstyle.version}</version>
```
xvrl: I _think_ you can remove all `<dependencies>` altogether here, given that they are already defined in the pluginManagement section. This should just inherit it.

ewencp:
ping. adjusted commit and pr title based on updated comment

PR #96, opened 2018-01-17
CLIENTS-835: Downgrade Avro from 1.8.2 to 1.8.1

ewencp:
1.8.2 contains a critical bug https://issues.apache.org/jira/browse/AVRO-2122
which customers are encountering. Although a fix is available
https://github.com/apache/avro/pull/276, it isn't merged yet and Avro releases
are very infrequent. The bug is not present in 1.8.1 and of the
fixes/improvements from 1.8.1 to 1.8.2 we are pretty confident 1.8.1 is going
to be the better release to go with.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
     <properties>
         <!--jacoco prepare-agent needs argLine defined to set this variable. if we don't have a default set then maven gets angry.-->
         <argLine></argLine>
-        <avro.version>1.8.2</avro.version>
+        <avro.version>1.8.1</avro.version>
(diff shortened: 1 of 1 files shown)


ewencp:
In theory we've got everything downstream depending on this version spec now, but we'll want to verify in nightly packaging output that there aren't any stray jars with a different version.

mageshn:
LGTM. We need to backport this as well.

ewencp:
@mageshn oops, for some reason I had it in my head 4.0 was when we upgraded that. moved the commit back to 3.3.x and changed base

PR #99, opened 2018-01-26
CC-1444: Use current thread's classloader if available

wicknicks:
Fixes a bunch of system tests failures. 

### DIFF
--- a/config/src/main/java/io/confluent/common/config/ConfigDef.java
+++ b/config/src/main/java/io/confluent/common/config/ConfigDef.java
@@ -361,7 +361,11 @@ public class ConfigDef {
           if (value instanceof Class) {
             return (Class<?>) value;
           } else if (value instanceof String) {
-            return Class.forName(trimmed);
+            if (Thread.currentThread().getContextClassLoader() != null) {
+              return Class.forName(trimmed, true, Thread.currentThread().getContextClassLoader());
+            } else {
+              return Class.forName(trimmed);
+            }
           } else {
             throw new ConfigException(name, value, "Expected a Class instance or class name.");
           }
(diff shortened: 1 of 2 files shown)


config/src/main/java/io/confluent/common/config/ConfigDef.java
```diff
@@ -361,7 +361,11 @@ else if (value instanceof String)
           if (value instanceof Class) {
             return (Class<?>) value;
           } else if (value instanceof String) {
-            return Class.forName(trimmed);
+            if (Thread.currentThread().getContextClassLoader() != null) {
```
mageshn: This should certainly fix the issue. But I was wondering if it made more sense to symlink the confluent-common into kafka-serde-tools. This would mean that confluent-common is available as part of the plugin-path and hence same class loader as serializers & converters. This should also be fixed in our packaging.

wicknicks: Yes, that is an alternative solution. We can create a softlink to confluent-common wherever we create links to storage-common. I just checked it, and that fixes the problem too.

mageshn: @kkonstantine will be great to hear your thoughts on this before we purse the above solution.

wicknicks:
@hachikuji: I added a test case. 

@kkonstantine, @mageshn: thanks! I'll merge this PR, and then fix the other repos in a future PR.

mageshn:
@wicknicks LGTM. May be we can merge this PR after we put in the ST fix. Just to make sure we know it works even without this fix :).

kkonstantine:
LGTM. IMO this fix is orthogonal to system tests and to test any functionality before this change we can always pin the branch for common to point to a specific commit in the system test branch builder.

kkonstantine:
@wicknicks I missed to comment on the target branch. Can you backport (merge) this fix to `3.3.x` and `4.0.x` as well?

wicknicks:
@kkonstantine: do we need to? the bug only affects master branch. The `TopicNameStrategy` and other implementations were merged to master.

PR #100, opened 2018-02-13
MINOR: Add copy constructor for ConfigDef

C0urante:
The `ConfigDef` in Kafka already has support for this; adding it to enable subclassing of `ConfigDef` while still playing nicely with our `rest-utils`, specifically the idiom of creating a `ConfigDef` by starting with the `RestConfig.baseConfigDef()` method and chaining `define()` calls from there.

### DIFF
--- a/config/src/main/java/io/confluent/common/config/ConfigDef.java
+++ b/config/src/main/java/io/confluent/common/config/ConfigDef.java
@@ -68,7 +68,7 @@ import io.confluent.common.config.types.Password;
  */
 public class ConfigDef {
 
-  private static final Object NO_DEFAULT_VALUE = new Object();
+  protected static final Object NO_DEFAULT_VALUE = new Object();
 
   private final Map<String, ConfigKey> configKeys;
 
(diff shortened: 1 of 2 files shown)


config/src/main/java/io/confluent/common/config/ConfigDef.java
```diff
@@ -68,9 +68,17 @@
...
+  protected static final Object NO_DEFAULT_VALUE = new Object();
 
-  private final Map<String, ConfigKey> configKeys = new HashMap<String, ConfigKey>();
+  private final Map<String, ConfigKey> configKeys;
+
+  public ConfigDef() {
+    configKeys = new HashMap<>();
+  }
+
+  public ConfigDef(ConfigDef baseConfigDef) {
```
mhowlett: i wouldn't use the word `base` due to the potential confusion with inheritance. I'd call the parameter `configDef`.

mhowlett:
LGTM

C0urante:
Thanks @mhowlett!

PR #101, opened 2018-02-23
MINOR: Update copyrights to not look like Javadoc style comments

wicknicks:

### DIFF
--- a/config/src/main/java/io/confluent/common/config/AbstractConfig.java
+++ b/config/src/main/java/io/confluent/common/config/AbstractConfig.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2015 Confluent Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,7 +14,7 @@
  * limitations under the License.
  **/
 
-/**
+/*
  * Original license:
  *
  * Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements. See the NOTICE
--- a/config/src/main/java/io/confluent/common/config/ConfigDef.java
+++ b/config/src/main/java/io/confluent/common/config/ConfigDef.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2015 Confluent Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,7 +14,7 @@
  * limitations under the License.
  **/
 
-/**
+/*
  * Original license:
  *
  * Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements. See the NOTICE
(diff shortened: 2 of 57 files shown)


ewencp:
@wicknicks What was the conclusion on the checkstyle regex missing this? And what is the fallout across all our repos? Are we going to end up having a similar PR across all repos? These aren't lines where history is important for blame purposes, so I'm fine changing this, but want to make sure a) we're going adjust checkstyle to catch anything like this in the future and b) we actually resolve this for the entire platform.

wicknicks:
@ewencp: I haven't changed the Checkstyle Regex. If we fix it here, the builds across a lot of repos will break. So we can migrate the repos one at a time, and hopefully catch new files having the wrong style in the meantime. Once we have a good majority of repos migrated to the new style, we can switch the regex. Maybe we have this only in the `common` repo, but I found some files with two Copyrights. 

For example, [in this file](https://github.com/confluentinc/common/blob/master/config/src/main/java/io/confluent/common/config/AbstractConfig.java#L17). The regex will not find those cases.

The only problem I see here, is to decide whether we want to fix this in all branches or only `master`. If we fix it in only `master`, then we run the risk of having merge conflicts while running a `pint merge`. It is a bit more work to do this for every branch, but possibly scriptable :-).

Yes, we would need a PR for every repo. There is no other way to do this (IIUC). 

I started this PR to mostly start a discussion (and get things rolling). If you think there is a better place to do this, let me know. Thought it was on overkill for one-pager. Maybe announce it in a Slack channel?

ewencp:
@wicknicks Fixing globally is fine, just wanted to make sure someone was actually signed up to do the follow up :) Not a complicated task, but also not a small one. It would also be good to have the plan in place where the change to the checkstyle rules will cause failures, but we already have fixes in place across the board to ensure we won't see those failures.

re: regex, we should probably try to understand why it is/isn't catching this currently. checkstyle docs do mention line-by-line evaluation, so possibly we have a bug in the checkstyle test since the regex clearly assumes multiple lines. I assumed this worked because there are suppressions for blueway which uses a different comment style, but maybe that was incorrect.

re: which branches, once you've done one merge from the patch onto a particular branch, the rest should not change for these copyright comments -- any conflicts would have been resolved. so we probably can just do it on master. that said, I by default try to figure out how far back we can backport fixes and target patches there & then merge up through newer release branches. i think the risk/cost in this case is for new files -- if you do this against, say, a 3.3.x branch, if new files were added they will be missed in the patch and require separate follow-up. not the end of the world, but possibly a little more messy.

wicknicks:
yeah, it can get a bit messy. although after thinking about it, I think it's better to merge all the main branches with these changes (it's 5 PRs and then a merge across branches). that way we don't worry about merge conflicts again.

I'll look into the regex too. will keep you updated.

ewencp:
@wicknicks Is this still worth pursuing? Can we either update and get it committed or close?

wicknicks:
probably not worth the effort. let's close it for now. thanks for the follow up, @ewencp.

PR #102, opened 2018-03-06
Define Jackson version centrally for use across the platform

ewencp:


ewencp:
See also https://github.com/confluentinc/rest-utils/pull/89, just moving this more centrally so you don't have to rely on rest-utils to get this.

PR #107, opened 2018-03-09
Bump Kafka dependency to 1.0.2-SNAPSHOT since 1.0.1 has been released

ewencp:


ewencp:
This currently blocks the kafka-1.0 job from succeeding since the downstream build won't work.

pom.xml
```diff
@@ -52,7 +52,7 @@
         <jackson.version>2.9.4</jackson.version>
         <jacoco-maven-plugin.version>0.7.9</jacoco-maven-plugin.version>
         <junit.version>4.12</junit.version>
-        <kafka.version>1.0.1-SNAPSHOT</kafka.version>
+        <kafka.version>1.0.2-SNAPSHOT</kafka.version>
```
maxzheng: Need to update https://github.com/confluentinc/common-tools/blob/master/confluent/config/dev/versions.json#L26

ewencp: I can file a PR for that. I guess we shouldn't actually merge this until 4.0.1 goes out anyway, we're just going to be stuck w/ broken kafka-1.0 builds until that happens.

maxzheng:
@ewencp FYI, for future changes like this, I believe the easier way is to update versions.json and then run `tbr bump-deps` and `pint merge` as there are other components that would need to be updated as well -- see full list of steps at https://confluentinc.atlassian.net/wiki/spaces/Engineering/pages/507838927/Public+Final+Release#PublicFinalRelease-STEP5:Bumpversionstopreparefornextrelease

PR #108, opened 2018-03-21
utility-belt should not override the jackson version defined in the parent pom

ewencp:


ewencp:
@aayars rebuilding to get to a clean build. i've seen that flakiness in other runs as well. i can't even get that set of tests to pass locally, so they could probably use some work...

PR #111, opened 2018-04-17
pip.req is not needed

aayars:


aayars:
Retargeting to 4.1.x.

PR #113, opened 2018-04-19
Inherit jackson version from parent pom instead of explicitly specifying in utility belt

ewencp:


ewencp:
Failure seems due to some timeout issue in the docker tests, but I think it's an error with infra and not a real error, retrying.

PR #117, opened 2018-05-02
Update versions to support spring boot

ryan-barker-zefr:
There will be a corresponding pull request on common-docker and another on rest-utils

This is required to utilize the rest-utils for unit tests in a spring boot application

https://github.com/ZEFR-INC/kafka-kstreams-kotlin-example-app

The basic issue is that spring boot requires newer versions of jetty and jersey while confluent is stuck on old versions.  Relatively minor changes are required to make this work, but it does require java 8 per comments (not tested). Considering java 7 has been end of life since April 2015 https://www.java.com/en/download/faq/java_7.xml I think this is reasonable.

Please feel free to fork the kotlin example app to include as a sample kstream app for kotlin. It's only non-public dependency is some compiled avro files that could be easily switched to some other avro files.

### DIFF
--- a/build-tools/pom.xml
+++ b/build-tools/pom.xml
@@ -20,6 +20,6 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>io.confluent</groupId>
     <artifactId>build-tools</artifactId>
-    <version>4.0.2-SNAPSHOT</version>
+    <version>4.0.1-ZEFR-1</version>
     <name>Build Tools</name>
 </project>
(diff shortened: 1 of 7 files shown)


ryan-barker-zefr:
Fixes https://github.com/confluentinc/rest-utils/issues/26

ryan-barker-zefr:
This was necessary to build the rest utils project due to parent pom being
in another repo. I agree this change does nothing for you.

The only real change was an upgrade of jetty and jersey versions for rest
utils.

xvrl:
closing, since 4.x is no longer supported

PR #123, opened 2018-05-16
CC-1975: Increase PermGen for surefire plugin

wicknicks:
This is to help downstream repos which need a larger PermGen for their tests. This will enable us to increase perm gen for kafka-connect-jdbc tests which are failing on Jenkins right now. Related PR: https://github.com/confluentinc/kafka-connect-jdbc/pull/413.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,6 @@
         <maven-enforcer-plugin.version>3.0.0-M1</maven-enforcer-plugin.version>
         <maven-jar-plugin.version>3.0.2</maven-jar-plugin.version>
         <maven-surefire-plugin.version>2.21.0</maven-surefire-plugin.version>
-        <maven-surefire-plugin.vmopts>-XX:MaxPermSize=128m</maven-surefire-plugin.vmopts>
         <mockito-all.version>1.10.19</mockito-all.version>
         <os-maven-plugin.version>1.5.0.Final</os-maven-plugin.version>
         <slf4j-api.version>${slf4j.version}</slf4j-api.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -456,7 +457,7 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
-                    <argLine>@{argLine} -Djava.awt.headless=true</argLine>
+                    <argLine>@{argLine} -Djava.awt.headless=true ${maven-surefire-plugin.vmopts}</argLine>
```
maxzheng: How about just setting it to 128mb as the default for everything? It's not much difference and avoids the need to set this individually for different repos that might come up as a problem later.

wicknicks: fixed above. thanks!

pom.xml
```diff
@@ -64,6 +64,7 @@
         <maven-enforcer-plugin.version>3.0.0-M1</maven-enforcer-plugin.version>
         <maven-jar-plugin.version>3.0.2</maven-jar-plugin.version>
         <maven-surefire-plugin.version>2.21.0</maven-surefire-plugin.version>
+        <maven-surefire-plugin.vmopts>-XX:MaxPermSize=128m</maven-surefire-plugin.vmopts>
```
maxzheng: I would also remove maven-surefire-plugin.vmopts and just hardcode to start. We can always add it later if needed.

ijuma:
This option has been removed in recent JDKs, so it would be nice to only enable it if Java 7 is being used.

wicknicks:
Thanks, @ijuma! Should we just revert this commit when we switch to Java 8? I am not sure we can do a conditional property for the surefire plugin. AFAIK, this should not hurt Java 8 VMs.

PR #127, opened 2018-05-22
ST-304: Update source/target/release to 8

ijuma:
- Introduce java.version variable.
- Remove MaxPermSize.


ijuma:
Followed up on most repos, asked @kkonstantine to do Connect as the compiler settings also needed to be updated.

PR #128, opened 2018-05-26
CC-2013: Use ClassLoader before trying to use ContextClassLoader

wicknicks:

### DIFF
--- a/config/src/main/java/io/confluent/common/config/ConfigDef.java
+++ b/config/src/main/java/io/confluent/common/config/ConfigDef.java
@@ -30,6 +30,9 @@
  */
 package io.confluent.common.config;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -67,6 +70,8 @@ import io.confluent.common.config.types.Password;
  * some additional functionality for accessing configs.
  */
 public class ConfigDef {
+
+  private static final Logger log = LoggerFactory.getLogger(ConfigDef.class);
 
   private static final Object NO_DEFAULT_VALUE = new Object();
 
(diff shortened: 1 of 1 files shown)


config/src/main/java/io/confluent/common/config/ConfigDef.java
```diff
@@ -361,10 +366,14 @@ else if (value instanceof String)
...
-              return Class.forName(trimmed, true, Thread.currentThread().getContextClassLoader());
-            } else {
+            try {
               return Class.forName(trimmed);
+            } catch (ClassNotFoundException e) {
+              log.error("Could not find class " + trimmed, e);
+              if (Thread.currentThread().getContextClassLoader() != null) {
+                log.error("Reattempting with context class loader " + trimmed, e);
+                return Class.forName(trimmed, true, Thread.currentThread().getContextClassLoader());
+              }
```
wicknicks: @mageshn, @kkonstantine does this change make sense to you? It use the default classloader (the one which originally created this class), before using the ContextClassLoader to instantiate classes. This is going to impact how Connect's plugin classloader works. If this change is incorrect, we need to rethink how we did https://github.com/confluentinc/schema-registry/commit/d8fd3bf6feaeb49433e8474e284b6cc0aabe713c (or how AvroConverter's dependencies are laid out in CP).

mageshn:
@wicknicks The right thing would be to fix the System Tests and if required the packaging as well. The requirements for the plugin class loader is simple - have all dependencies in the plugin path

wicknicks:
A run of the tests without the classpath issues: http://confluent-platform-branch-builder-system-test-results.s3-us-west-2.amazonaws.com/2018-05-26--002.1527364074--wicknicks--HEAD--b5697d9/report.html

wicknicks:
@kkonstantine: I'll probably close this PR, since we are mostly likely going to change how packaging is done (which seems to be the right approach to me). @mageshn?

norwood:
@wicknicks @mageshn ping

wicknicks:
Problem was fixed by this PR: https://github.com/confluentinc/common/pull/133

PR #129, opened 2018-05-26
Upgrade Avro to 1.8.2

ijuma:
Some projects that depend on common are using Avro 1.8.2
already and this causes issues in the tarball
distribution due to multiple versions in the classpath.
Thanks to @maxzheng for diagnosing a system test
failure due to this issue.


cricket007:
What about the "issue" mentioned in https://github.com/confluentinc/common/pull/96 ?

ijuma:
Good point. Given that, it seems like the right solution is to downgrade the other projects too. I will submit a PR to revert this and to include a comment staying that 1.8.2 should be skipped.

PR #131, opened 2018-06-07
Add license as upstream and remove kafka

maxzheng:
kafka-trunk is a freestyle job and will not work with upstreamProjects as we append the branch after the job name.

We could fix upstreamProjects to work, but probably better without as having to maintain the branch mapping is not going to happen, which is the case based on current values used. So just drop it until we have better ways to map the branches automatically.

Also change to use dockerfile as docker_oraclejdk8 is deprecated

Replay: https://jenkins.confluent.io/job/confluentinc/job/common/job/4.1.x/151/


ijuma:
Description sounds good although I'm not familiar with the specifics. :)

PR #133, opened 2018-06-18
Fix to use caller class loader only.

mageshn:
The ability to use ThreadContext classloader was added to fix a System test for Connect. The fix is not helping now in both packaging and system tests. Reverting the fix to go back to the original implementation. If we need to use the current class loader, Connect should possibly set the current class loader for all the plugins.


mageshn:
@kkonstantine Here is the branch builder logs https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/927/console
https://s3.console.aws.amazon.com/s3/buckets/confluent-platform-branch-builder-system-test-results/2018-06-19--002.1529460714--confluentinc--HEAD--deca23e/?region=us-west-2&tab=overview

The issue that we are trying to fix is not happening anymore. But the job is failing for other reasons.

PR #134, opened 2018-06-29
Easymock Java 10 support + fix maven-site-plugin issues

xvrl:


ijuma:
Maven site issue affecting the PR builder.

xvrl:
and we're green

assembly-plugin-boilerplate/pom.xml
```diff
@@ -11,17 +11,20 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
     <name>${project.artifactId}</name>
     <description>Maven Assembly Plugin boilerplate for Dockerized projects</description>
 
-    <properties>
-        <maven-assembly.version>3.0.0</maven-assembly.version>
```
ijuma: My understanding is that we are trying to centralize versions as properties. cc @ewencp

ewencp: where there is a chance of inheritance or reuse, yes. it makes it easier than digging through 80 projects to find all the places we need to upgrade something, and if we need to override it in a specific project, we still have that option.

xvrl: It's always been my understanding that the pluginManagement (or dependencyManagement) section plays that role. I don't see the added benefit of properties unless they're shared among multiple artifacts (e.g. jackson). For all our artifacts the version will only ever be defined once in the management section

xvrl: Just to clarify, once you define the version in the dependency-management section, every project that inherits from common will automatically inherit that version as long as it does not specify a version in the dependency section. Same applies to plugin-management.
In a way the dependency management section goes even further than a property, since you can also define other configurations that will be inherited in sub-projects.

PR #136, opened 2018-07-01
Set version of maven-site and maven-project-info-reports to ensure compatibility

ijuma:
This is a selective backport of the following commit that was merged to
5.0.x and master:

https://github.com/confluentinc/common/commit/d5c973a2c766b61a8ff789dc0768ae005df63209

PR #137, opened 2018-07-03
MINOR: Corrected version of Powermock Mockito artifact

rhauch:
The correct artifact is `org.powermock:powermock-api-mockito2` for version `2.0.0-beta.5`. See the [Maven Central listing](https://search.maven.org/#search%7Cga%7C1%7Cpowermock-api-mockito).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -275,7 +275,7 @@
             </dependency>
             <dependency>
                 <groupId>org.powermock</groupId>
-                <artifactId>powermock-api-mockito</artifactId>
+                <artifactId>powermock-api-mockito2</artifactId>
                 <version>${powermock.version}</version>
                 <scope>test</scope>
             </dependency>


rhauch:
I found this when dealing with other errors/warnings from the `maven-plugin-info-reports` plugin, which was upgraded and pinned recently. However, this issue was introduced in 4.1.x when we upgraded to 2.0.0-beta.5

pom.xml
```diff
@@ -268,7 +268,7 @@
             </dependency>
             <dependency>
                 <groupId>org.powermock</groupId>
-                <artifactId>powermock-api-mockito</artifactId>
+                <artifactId>powermock-api-mockito2</artifactId>
```
xvrl: if this worked before then that means none of our dependencies actually reference mockito2, or if they do they probably override the version somehow

xvrl:
@rhauch your build is failing, are you sure you rebased on the latest version of this branch? It shouldn't  fail the way if did unless you don't have the change from #136

pom.xml
rhauch: Local developers might have it cached, but as long as our builds don't catch any artifacts in a local Maven repo, then it's probably a signal that it's not used or overridden. Either way, I wanted to correct this without necessarily causing more havoc downstream. It's been in the common since 3.3.x, and there are a handful of sister artifacts.

Maybe PowerMock is not used with Mockito? I've used Mockito, but probably not PowerMock at the same time, and Connect uses PowerMock with EasyMock (not affected by this artifact). It is overridden in at least [one repo](https://github.com/confluentinc/kafka-connect-storage-cloud/blob/master/pom.xml#L137-L142).

@xvrl, are you suggestion we remove the definition?

xvrl: I suggest we figure out if we have any repos actually using it, if no, then let's remove it, if yes, let's make sure they don't override it

rhauch:
@xvrl corrected the PR with the correct rebase.

norwood:
is this still needed?

@rhauch

rhauch:
@norwood, yes, it still is needed IMO since the referenced artifact in the existing POM does not exist. Alternatively we could remove the artifact from this POM since it's obviously not causing any problems (other than warnings during the build) -- several of the Connect repos are bringing in the correct artifact on their own.

Apologies for the late response -- been on holiday since CKO.

PR #138, opened 2018-07-03
MINOR: Disabled dependency location feature of maven-plugin-info-reports

rhauch:
We recently pinned the version of the `maven-plugin-info-reports` plugin to 3.7.1, and this resulted in some new warnings about blacklisting repositories. For example:
```
[WARNING] The repository url 'https://nexus.confluent.io/repository/maven-public' is invalid - Repository 'confluent-nexus-central' will be blacklisted.
```
Apparently disabling the dependency locations check will eliminate these warnings.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -427,6 +427,13 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-project-info-reports-plugin</artifactId>
                     <version>${maven-project-info-reports-plugin.version}</version>
+                    <configuration>
+                        <!-- 
+                        Disable dependency locations for latest maven-plugin-info-reports to eliminate blacklisting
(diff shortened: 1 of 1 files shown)


rhauch:
This blacklisting warning did not occur on older branches.

pom.xml
```diff
@@ -427,6 +427,10 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-project-info-reports-plugin</artifactId>
                     <version>${maven-project-info-reports-plugin.version}</version>
+                    <configuration>
+                        <!-- Disable dependency locations for latest maven-plugin-info-reports to eliminate blacklisting -->
```
xvrl: maybe include the warning, since it may not be clear what blacklisting means

rhauch:
@xvrl mind taking another pass?

norwood:
@rhauch ping

PR #141, opened 2018-07-17
Add FinalParametersCheck as an optional checkstyle check.

ewencp:
Downstream projects can enable the check by setting the severity higher (e.g. warning or error) with the `checkstyle.finalparameter.severity` property using either `propertyExpansion` or `propertiesLocation` settings for the checkstyle plugin (https://maven.apache.org/plugins/maven-checkstyle-plugin/examples/custom-property-expansion.html).

### DIFF
--- a/build-tools/src/main/resources/checkstyle/checkstyle.xml
+++ b/build-tools/src/main/resources/checkstyle/checkstyle.xml
@@ -221,9 +221,9 @@
         <module name="JavaNCSS" />
         <module name="NPathComplexity"/>
 
-        <module name="FinalParametersCheck">
+        <module name="FinalParameters">
             <property name="severity"
-                      value="${checkstyle.finalparameter.severity}"
+                      value="${checkstyle.finalparameters.severity}"
                       default="ignore"/>
         </module>
     </module>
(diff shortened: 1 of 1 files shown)


build-tools/src/main/resources/checkstyle/checkstyle.xml
```diff
@@ -220,5 +220,11 @@
         <module name="CyclomaticComplexity"/>
         <module name="JavaNCSS" />
         <module name="NPathComplexity"/>
+
+        <module name="FinalParametersCheck">
```
big-andy-coates: nit: Switch `FinalParamtersCheck` to `FinalParameters`, as none of the other checks postfix with `Check`

build-tools/src/main/resources/checkstyle/checkstyle.xml
```diff
@@ -220,5 +220,11 @@
...
+            <property name="severity"
+                      value="${checkstyle.finalparameter.severity}"
```
big-andy-coates: nit: `checkstyle.finalparameters.severity`, (not the 's' on the end of parameters)

PR #149, opened 2018-07-19
Add opt-in RedundantThrows check.

big-andy-coates:
(And correct the default on the `RedundantImport` check.


big-andy-coates:
Meh, this check was removed from CheckStyle: https://github.com/checkstyle/checkstyle/issues/473

PR #151, opened 2018-07-27
Add ability to suppress checkstyle rules using comments.

big-andy-coates:
This, in general, works better with IDEs that a suppression file, and puts the suppression with the code.

E.g. to suppress a one or more rules:

```
// CHECKSTYLE_RULES.OFF:  ClassDataAbstractionCouplingCheck|CyclomaticComplexityCheck - This class is complex because blah blah blah
public class SomeJustifiablyComplexClass {
 ...
}
// CHECKSTYLE_RULES.ON:  ClassDataAbstractionCouplingCheck|CyclomaticComplexityCheck
```

E.g. to turn of checkstyle all together:

```
// CHECKSTYLE:OFF - Checkstyle is turned off because this code is generated by blah.
... some generated code.
// CHECKSTYLE:ON
```

### DIFF
--- a/build-tools/src/main/resources/checkstyle/checkstyle.xml
+++ b/build-tools/src/main/resources/checkstyle/checkstyle.xml
@@ -41,6 +41,19 @@
     </module>
 
     <module name="TreeWalker">
+        <!-- Generic 'turn all checkstyle off' comment filter -->
+        <module name="SuppressionCommentFilter">
+            <property name="offCommentFormat" value="CHECKSTYLE:OFF"/>
+            <property name="onCommentFormat" value="CHECKSTYLE:ON"/>
+        </module>
+
+        <!-- Specific 'turn list of checkstyle rules off' comment filter -->
+        <module name="SuppressionCommentFilter">
+            <property name="offCommentFormat" value="CHECKSTYLE_RULES.OFF:\s+([\w\|]+)"/>
+            <property name="onCommentFormat" value="CHECKSTYLE_RULES.ON:\s+([\w\|]+)"/>
+            <property name="checkFormat" value="$1"/>
+        </module>
+
         <module name="OuterTypeFilename"/>
         <module name="IllegalTokenText">
             <property name="tokens" value="STRING_LITERAL, CHAR_LITERAL"/>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <slf4j.version>1.7.25</slf4j.version>
         <zkclient.version>0.10</zkclient.version>
         <zookeeper.version>3.4.13</zookeeper.version>
-        <checkstyle.version>6.19</checkstyle.version>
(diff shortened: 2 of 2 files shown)


ewencp:
@norwood do you remember why we went with only suppression files? i think i vaguely recall someone not liking inline toggling, but i can't remember the reasoning. possibly that it makes it too easy to switch things off and can be missed in review more easily, so people would sneak things by that should actually be fixed?

big-andy-coates:
The reason I'm adding it is that it works better with things like the CheckSyleIDEA plugin for IntelliJ.  The suppression file isn't picked up, but the inline code comments are. Which means you can run CheckStyle inside the IDE without then having to _ignore_ the suppressed errors.  Without this, there's just too much noise.

Having said that... the build failed on this with error 

> TreeWalker is not allowed as a parent of SuppressionCommentFilter

Which may also be a reason why you didn't go with suppression comments. I've no idea why its failing! It works locally.  Is the build doing anything special in this regard?  Maybe updating the checkstyle version will fix this, but I really don't know.

big-andy-coates:
Worked it out - weirdness going on locally, but got it to fail locally with a restart.

So the IDE plugins use later versions of Checkstyle than the build and each requires the SuppressionCommentFilter in a different place.  Hence, I've updated the Checkstyle version used in the build.  My testing with KSQL has shown this doesn't cause any build failures. (Moving up to 8.6 does, due to missing `<p>` in javadocs). Other projects may vary, but hopefully this isn't a build breaking change.

big-andy-coates:
Feel free to merge this if/when you approve, as I'll be offline and it would be handy to come in Monday and have a snapshot build already. :D

big-andy-coates:
@Ewen, @norwood is not in today. He didn't seem to have any objections about comment based checkstyle handling when he replied on slack...

big-andy-coates:
Hey @ewencp,

> Well, the change itself LGTM as long as we've made sure nobody really had objections. I may just be misremembering.

I'm not sure how I can make sure nobody really has objections, short of mailing the entire dev team. At which point, I can guarantee objections and probably another flame war ;)

But looking at this from a practical standpoint, we already allow suppressions via a file, so why not by comments too? Especially, given that the latter works well with the tooling and the former doesn't.

> For the checkstyle version, did this just get updated because we were stuck on 6.19 for jdk7?

Not sure why we've  been stuck on 6.19, could of been a jdk version thing, but updating seems to be sensible. Otherwise the cost of updating, as I'm sure you're aware, just keeps growing.

This isn't to the latest, but the latest generates loads of errors due to some java doc issues. We'll need to work out how to turn that off before we update to the latest.

norwood:
👍 on comment suppressions.

i think i actually prefer these. its much easier to ensure that you are only making exceptions in one place in the file rather than for the whole file like what usually ends up happening with the suppressions.xml

ijuma:
And we had to update a bunch of code in AK after the upgrade to the newer version.

PR #152, opened 2018-07-27
slf4j-log4j12 in test scope causes downstream issues

xvrl:
Having sl4fj-log4j12 as an explicit test-scoped dependency causes subtle
issues for downstream projects that use common-pom as a parent.

Projects that need log output during test runs, should either explicitly
declare a dependency on slf4j-log4j12 (either test or runtime scoped,
depending on whether that project should ship a logging implementation).

Bear with me, but here's an example of what can happen:
- project A uses common-parent and has sl4fj-log4j12 as a runtime dependency
- project B also uses common-parent and has project A as dependency

You would expect that project B would now also have slf4j-log4j12 as a
runtime dependency. Except, because common-parent has an explicit
depenency in test-scope, this now overrides the scope defined in A.
As a result, project B now doesn't produce any logs.

This means any project using common-parent that needs a logging
implementation, also has to specifically redefine a runtime scope
dependency on sl4fj-log4j12 to override the common pom, even if that
runtime dependency should have been pulled in transitively.

This change now requires projects to explicitly declare a test-scope
dependency on a slf4j logging implementation if they need
logging output during testing but don't want to ship a logging
implementation.

One alternative I see to this change would be to declare a common
test-scoped dependency on slf4j-simple, so tests still produce logs
without conflicting with any logging implementation we might want to
ship (assuming of course we never want to ship slf4j-simple)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -327,11 +327,6 @@
             <type>zip</type>
             <scope>provided</scope>
         </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-log4j12</artifactId>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 
     <build>


ewencp:
In your scenario, shouldn't project B be explicitly specifying the runtime dependency on slf4j-log4j12 anyway? Given our historical issues with multiple projects (Kafka, Zkclient, etc) incorrectly including log4j with *libraries* rather than just alongside *services*, it seems like having the test scoped dependency is actually *good* for us since it forces us to explicitly add log4j at each service.

(Only partly related, but also it seems like in your scenario project A either needs to be refactored to provide a library component that doesn't have log4j or any "service" dependencies included, which is what downstreams should use, and use a different module for the actual service in A. Either that, or A should move the dependency on log4j into only the modules that need it.)

xvrl:
@ewencp removing the dependency here will also force each service to explicitly add log4j, so I don't see the benefit of keeping it.

The problem I am having is that sometimes you just want project B to depend on project A for packaging purposes. In my case, project B creates a docker image that include A along a few other things. A is a service so it correctly ships log4j. I now expect B to also have everything that A comes with, because B is just a repackaged version of A. I shouldn't have to redeclare my log4j dependency.

ewencp:
Can we make this concrete and name what A and B are in the case you're running into this?

This is only targeted at master, so I guess not an issue, necessarily, but https://issues.apache.org/jira/browse/KAFKA-6317 wasn't fixed until pretty recently in 1.1.0, I'm not sure how long it had appeared. I believe that's why we end up having to do https://github.com/confluentinc/common/blob/master/pom.xml#L201.

xvrl:
@ewencp it's hard to prevent people pulling in all the random libraries they might only need for testing. Maybe there is a better way to flag those things using the maven enforcer plugin in the relevant packages that should not ship log4j

PR #153, opened 2018-07-30
Move utility-belt out, to eliminate Kafka build dep.

aayars:
Accompanies PR https://github.com/confluentinc/common-docker/pull/14


Jenkinsfile
```diff
@@ -1,10 +1,7 @@
...
 
-dockerfile {
-    upstreamProjects = ['confluentinc/confluent-docker-utils', 'confluentinc/license-file-generator']
-    dockerRegistry = '368821881613.dkr.ecr.us-west-2.amazonaws.com/'
-    dockerRepos = ['confluentinc/cp-base']
+common {
+    upstreamProjects = ['confluentinc/license-file-generator']
     slackChannel = '#kafka-core-eng'
     nodeLabel = 'docker-oraclejdk7'
-    dockerPush = true
```
maxzheng: Generally, this PR looks like. The only concern is the impact of this change on downstream projects. I kinda asked about this in ur other PR, but what are the impacts to downstream projects and follow-up changes once this goes in since they all inherit from common atm but not common-docker?

aayars: Child projects can continue using confluentinc/common for a parent, which will continue to provide the assembly boilerplate and Dockerfile support. This should be a pretty transparent change downstream, since almost nothing depends on the utility belt classes directly.

One missing piece is getting the new version built automatically at release time, so the docker images can pull in the release version.

PR #157, opened 2018-07-31
Bump AK to 1.1.1-cp1

mohnishbasha:
Bump AK to 1.1.1-cp1
* AK 1.1.1-snapshot doesnt exists in the repo.

Currently,  kafka-1.1/ packaging-4.1.x are failing due to this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
         <jackson.version>2.9.6</jackson.version>
         <jacoco-maven-plugin.version>0.7.9</jacoco-maven-plugin.version>
         <junit.version>4.12</junit.version>
-        <kafka.version>1.1.1-SNAPSHOT</kafka.version>
+        <kafka.version>1.1.1-cp1</kafka.version>
         <kafka.scala.version>2.11</kafka.scala.version>
         <licenses.version>${project.version}</licenses.version>
         <maven-assembly.version>3.0.0</maven-assembly.version>


pom.xml
```diff
@@ -55,7 +55,7 @@
         <jackson.version>2.9.6</jackson.version>
         <jacoco-maven-plugin.version>0.7.9</jacoco-maven-plugin.version>
         <junit.version>4.12</junit.version>
-        <kafka.version>1.1.1-SNAPSHOT</kafka.version>
+        <kafka.version>1.1.1-cp1</kafka.version>
```
maxzheng: I don't think this will work as CP Kafka in 1.1 branch is producing a SNAPSHOT version. 
Also, any kafka.version changes should be done in common-tools and then use tbr. Ad-hoc changes like this will get reverted.

mohnishbasha: not sure, i understand it right. Are you suggesting to make changes in versions.json first for 4.1.x and then use as:

tbr bump-deps build 4.1.x --snapshot --no-with-confluent --with-kafka -e (dev, prod) PLATFORM --dry-run ?

maxzheng: 1. we can't set this to use -cp1 as that's a fixed version, and we don't bump versions per commit yet. 
2. if we do need to make a change to cp / kafka versions, they should always be done using tbr and not manually as pom.xml is controlled / supported by tbr.

mohnishbasha: 1. I understand that we gotta bump it all over.
2. tbr is the way to go, but should i be changing versions.json for 4.1.x to 1.1?

Here is the issue I am trying to address as the last step for 4.1.2-rc1:
packaging/4.1.x
kafka-1.1

Both are failing for a while since kafka 1.1.1-snapshot doesn't exists.

In versions.json, 4.1.x currently points to 1.1.2, however 1.1.2-snapshot doesn't exists.

maxzheng: AK is at 1.1.2-SNAPSHOT, but that won't be sync'ed until the sync job pass https://github.com/apache/kafka/blob/1.1/gradle.properties#L19

And the sync job won't pass as CP 4.1.x is still using AK 1.1.1-SNAPSHOT. https://jenkins.confluent.io/job/kafka-1.1/

You need to follow the steps in the Version Bump section: https://confluentinc.atlassian.net/wiki/spaces/Engineering/pages/507838927/Public+Final+Release#PublicFinalRelease-STEP5:Bumpversionstopreparefornextrelease

Specifically the one that tells you to uncheck the blocking to sync first, and then re-check later.

mohnishbasha: Yup, I went through those steps:
Was with the uncheck https://jenkins.confluent.io/job/kafka-1.1/335/
Rest all are failing with the check on.

I will give it a try again.

mohnishbasha: Tried it again with Re-check "Block until the triggered projects finish their builds" in the Kafka job. Downstream jobs still fails:
https://jenkins.confluent.io/job/test-cp-downstream-builds/3965/

maxzheng: Because common is still pinned to 1.1.1-SNAPSHOT: https://github.com/confluentinc/common/blob/4.1.x/pom.xml#L58

You need to run the `tbr bump-deps` so it sets the correct CP / AK version per versions.json per wiki

PR #159, opened 2018-08-08
update report plugin to avoid warnings with jdk10

xvrl:


xvrl:
@ijuma yes, but if you don't inherit from common you're already on your own :)

PR #160, opened 2018-08-09
avoid pulling in slf4j and log4j through commmon-utils

xvrl:


cmccabe:
What's the rationale?

xvrl:
@cmccabe when we ship a library we should not be imposing the logging backend, i.e. we should depend only on slf4j-api and let the user decide what logging implementation they are going to use.

stanislavkozlovski:
LGTM and the rationale is solid

cmccabe:
Thanks for the explanation.  LGTM.

PR #163, opened 2018-09-11
Switch from findbugs to spotbugs

ewencp:


ewencp:
findbugs is well known to be abandoned and unmaintained. spotbugs is just going to become the new standard.

this actually isn't a standalone change, but I needed to collect all the downstream changes as well. These are projects that needed build changes and we should coordinate merging them with this PR (their PR builds will fail until this is merged, if I have 👍 I can merge this, retrigger those PR builds, get a clean build and merge them):

https://github.com/confluentinc/kafka-rest/pull/467
https://github.com/confluentinc/kafka-connect-hdfs/pull/371
https://github.com/confluentinc/ksql/pull/1870

And this PR for jenkins-common that catches both spotbugs and findbugs xml reports should be merged before anything else since it still works with the old findbugs filenames: https://github.com/confluentinc/jenkins-common/pull/73

ewencp:
oh, by the way, for testing I just used a modified muckrake build that ensures we run spotbugs. that may not cover *quite* all our projects, but is enough that I don't think any other fallout would be difficult to fix.

pom.xml
```diff
@@ -51,8 +51,8 @@
         <confluent.version>5.0.1-SNAPSHOT</confluent.version>
         <easymock.version>3.6</easymock.version>
         <exec-maven-plugin.version>1.5.0</exec-maven-plugin.version>
-        <findbugs-maven-plugin.version>3.0.5</findbugs-maven-plugin.version>
-        <findbugs-annotations.version>3.0.1</findbugs-annotations.version>
+        <spotbugs-maven-plugin.version>3.1.6</spotbugs-maven-plugin.version>
+        <spotbugs-annotations.version>3.1.6</spotbugs-annotations.version>
```
ijuma: There was a regression in 3.1.6 that caused issue when building Kafka. 3.1.5 was fine. Do we want to be conservative here too?

pom.xml
```diff
@@ -637,11 +637,9 @@
             <reporting>
                 <plugins>
                     <plugin>
-                        <groupId>org.codehaus.mojo</groupId>
-                        <artifactId>findbugs-maven-plugin</artifactId>
+                        <groupId>com.github.spotbugs</groupId>
+                        <artifactId>spotbugs-maven-plugin</artifactId>
                         <configuration>
-                            <findbugsXmlOutput>true</findbugsXmlOutput>
-                            <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
```
ijuma: I assume these are no longer needed. Might be worth explaining why if others wonder in the future. :)

pom.xml
ewencp: What was the issue? as long as we have a working build, I don't think it matters to be on the newer version.

Worst case, we just reduce the version if we hit the same problem.

pom.xml
ewencp: It seems like spotbugs defaults are either different or these were superfluous anyway. Are you suggesting including a comment here? Seems a bit odd since once we're on spotbugs, nobody should care about findbugs anyway.

ewencp:
There's one more downstream that needs to be updated:

https://github.com/confluentinc/kafka-connect-replicator/pull/141

But in that case, it's a legit bug and not related to the build, so we can merge that one before anything else goes in.

pom.xml
ijuma: I just meant as a comment in the PR as you have done.

pom.xml
ijuma: It caused a stacktrace during the build. Since you used muckrake to test most projects, maybe that's fine. In any case, 3.1.7 will be out soon with a fix for the issue and Java 11 support, so hopefully we migrate to it before anyone can run into the problem while building locally.

ewencp:
Yeah, there's no rush here, so we can wait until it appears in maven and I'll update and re-verify against muckrake. Looks like it might be a couple of days based on previous releases. I'm not sure of spotbugs release process, but we may also want to wait for the updated release of the maven plugin, which I didn't see a 3.1.7 for when I checked earlier today.

ijuma:
I'd say merge.

xvrl:
@ewencp 3.1.7 is out now, you can update :)

ewencp:
@xvrl Can't quite yet, the main spotbugs jars were uploaded, but nothing's happened on the maven plugin which it looks like pins specifically to the same version of the spotbugs jar as the version of the plugin.

xvrl:
@ewencp according to this https://spotbugs.readthedocs.io/en/stable/maven.html you should be able to override the spotbugs version in the plugin config

xvrl:
@ewencp looks like the maven plugin is not ready for Java 11, that's why the author is waiting to bump the release: https://github.com/spotbugs/spotbugs-maven-plugin/issues/83#issuecomment-422974462

ijuma:
@xvrl I assume you're just sharing that as a FYI since it shouldn't matter for this merge right?

xvrl:
@ijuma yes, that was just in response to @ewencp 's comment about preferring not to have a separate version to track.

PR #166, opened 2018-10-10
Upgrade checkstyle and maven compiler plugin, enforce consistent EOF line endings

xvrl:
* Upgrade checkstyle and maven checkstyle plugin to latest versions
* Upgrade maven compiler plugin to support Java 10
* Enforce consistent newlines at end of files

### DIFF
--- a/build-tools/src/main/resources/checkstyle/checkstyle.xml
+++ b/build-tools/src/main/resources/checkstyle/checkstyle.xml
@@ -35,6 +35,8 @@
         <property name="eachLine" value="true"/>
     </module>
 
+    <module name="NewlineAtEndOfFile" />
+
     <!-- header -->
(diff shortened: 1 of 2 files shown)


ijuma:
Won't this cause a bunch of failures?

xvrl:
@ijuma we might as well fix them, shouldn't be too hard

xvrl:
sure it's debatable, as are some other whitespace formatting decisions, like an empty line after class declarations. As with everything it's about consistency. At the very least it avoids git warnings about missing newlines and makes diffs display more cleanly when only appending to a file.

ewencp:
@xvrl Can we close this out, or update and get it committed? It looks like we've got the same or newer versions now for all the version updates. The only thing currently missing from master is the newline setting for checkstyle.

PR #168, opened 2018-10-16
JDK11 support

xvrl:


xvrl:
These maven plugin versions should be safe to bump:
- Jacoco is a bug fix release
- Maven compiler changes are to require at least Java 1.7 and to default to source/target 1.6 instead of 1.5

Do we typically run full packaging when we bump those?

ewencp:
usually not packaging, especially since that's more of a pain. but pretty much anything in the common build i will run at least the muckrake build.sh (or ducker build) just to sanity check. doesn't validate everything since it doesn't run all tests and other stuff, but it's a decent sanity check

xvrl:
@ewencp @ijuma this will have to wait until we resolve downstream builds broken for other reasons.

PR #172, opened 2018-11-12
move shared dependencies to common

xvrl:


pom.xml
```diff
@@ -69,10 +70,12 @@
         <maven-surefire-plugin.version>2.21.0</maven-surefire-plugin.version>
         <maven-site-plugin.version>3.7.1</maven-site-plugin.version>
         <maven-project-info-reports-plugin.version>3.0.0</maven-project-info-reports-plugin.version>
+        <mockito.version>2.23.0</mockito.version>
         <mockito-all.version>1.10.19</mockito-all.version>
```
ijuma: Is anything still using Mockito 1.10?

pom.xml
```diff
@@ -231,6 +234,11 @@
                 <artifactId>slf4j-log4j12</artifactId>
                 <version>${slf4j.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.google.code.gson</groupId>
+                <artifactId>gson</artifactId>
+                <version>${gson.version}</version>
+            </dependency>
```
ijuma: Why is this under `kafka deps`?

pom.xml
xvrl: Yes we have some private repos that use it

pom.xml
xvrl: Fixed

PR #173, opened 2018-11-28
Add a structured logger and log4j1/2 extensions for logging to kafka

rodesai:
This patch adds interfaces for structured logging. Structured log messages
are represented as instances of connect SchemaAndValue, and would typically
contain objects of type Struct. The main interfaces to the structured logger
are a factory and the logger itself. The factory takes a prefix to prepend
to the name of each logger it creates.

The logger has functions for logging structured messages at various log levels.
The implementation just logs to slf4j using the parameters
<level>.("{}", structuredMsg). The underlying logging implementation can be
extended to handle structured msgs in more interesting ways (e.g. serializing
them as avro/json and writing to kafka).

The object passed to slf4j is of type SerializableSchemaAndValue, which
adds a toString message that returns the value as json. This way, the
messages are printed to the log (e.g. a file or console) in json by default.

The logger also has interfaces that accept Supplier<SchemaAndValue>, so users
can avoid constructing messages if the logger is disabled.

This patch also includes a module with some log4j2 extensions for formatting
structured messages as json or avro. This can be used in conjunction with the
KafkaAppender (which ships with log4j2) to log structured messages to kafka.


utils/src/main/java/io/confluent/common/logging/SerializableSchemaAndValue.java
```diff
@@ -0,0 +1,72 @@
...
+  public SchemaAndValue getMessage() {
+    return schemaAndValue;
+  }
+
+  private String serializeToString() {
+    final byte[] bytes =
+        converter.fromConnectData("", schemaAndValue.schema(), schemaAndValue.value());
+    try {
+      return objectMapper.writeValueAsString(
+          objectMapper.readValue(bytes, Object.class));
```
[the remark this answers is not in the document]
rodesai: This was changed in the latest revision. Its just `new String(converter.fromConnectData(...), UTF-8)` now.

utils/pom.xml
```diff
@@ -39,6 +39,14 @@
...
             <artifactId>slf4j-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>connect-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>connect-json</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
rodesai: Makes sense. I'll break this out into a separate module.

log4j2-extensions/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.11.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j-impl</artifactId>
+            <version>2.11.1</version>
```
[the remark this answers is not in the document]
rodesai: I'll just drop the dependency. We don't need the binding in this library.

utils/pom.xml
```diff
@@ -57,5 +65,24 @@
             <groupId>com.101tec</groupId>
             <artifactId>zkclient</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
```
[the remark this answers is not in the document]
rodesai: I moved the logging stuff into its own module, so I think this no longer applies.

log4j2-extensions/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>compile</scope>
```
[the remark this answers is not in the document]
rodesai: copypasta. fixed.

log4j2-extensions/src/main/java/io/confluent/common/logging/log4j2/StructuredLayout.java
```diff
@@ -0,0 +1,77 @@
...
+  public String getContentType() {
+    return "bytes";
+  }
+
+  public byte[] toSerializable(final LogEvent event) {
+    return toByteArray(event);
+  }
+
+  StructuredLayout(
+      final String topic,
```
xvrl: why do we need a topic name for structured logs?

rodesai: Its needed for the avro layout so it can compute a subject name to write the schema under. We should clean it up from this class though. Fixing that.

xvrl: yeah, it feels like a lot of connect-api specifics are leaking into the abstraction. It feels like the format / schema should just be implementation details, and the structured logging interface should just provide the necessary skeleton for us to plug in various implementations, whether that's json, avro, protobuf, it shouldn't really matter.

rodesai: That's pretty much how it's implemented.

The stuff under `common/logging/` is the logging interface. It has no details about the underlying format. It just has an interface that accepts SchemaAndValue as a structured log record, and an implementation that passes it along to the underlying logger. The object that's passed along (SerializableSchemaAndValue) does convert the msg to json, but that's just because some of the logging implementations (e.g. log4j1) force everything into String. There's also some builders there that help the various layouts for the various logging implementations generate the final log records with details about level, time, etc.

The layouts under log4j-extensions and log4j2-extensions is where the actual msg format is decided. This stuff is all pluggable via log4j1/log4j2 configuration.

logging/src/main/java/io/confluent/common/logging/LogRecordBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+  String FIELD_LOGGER = "logger";
+  String FIELD_LEVEL = "level";
+  String FIELD_TIME = "time";
+  String FIELD_MESSAGE = "message";
+
+  String NAMESPACE = "io.confluent.common.logging";
+
+  static SchemaBuilder baseSchemaBuilder() {
+    return SchemaBuilder.struct()
+        .name(NAMESPACE + "StructuredLogRecord")
```
vcrfxia: nit: either add a period between `NAMESPACE` and the name, or at the end of `NAMESPACE`? Everything else looks great. Sorry for the nit :)

rodesai: No worries - this is not a nit. Very important point in fact. Good catch!

PR #176, opened 2019-02-26
Move logging libs under logging group

rodesai:
This patch moves the logging libs under the group io.confluent.logging.
This way these artifacts don't get sucked up into the common-* wildcard
excludes in downstream packaging assemblies, e.g.:

https://github.com/confluentinc/ksql/blob/1ae1177b820109588413ea01b3f615cca746634e/ksql-package/src/assembly/package.xml#L82

### DIFF
--- a/log4j-extensions/pom.xml
+++ b/log4j-extensions/pom.xml
@@ -31,6 +31,7 @@
     </properties>
 
     <artifactId>confluent-log4j-extensions</artifactId>
+    <groupId>io.confluent.logging</groupId>
     <packaging>jar</packaging>
 
     <dependencies>
@@ -40,7 +41,7 @@
             <version>1.2.17</version>
         </dependency>
         <dependency>
-            <groupId>io.confluent</groupId>
+            <groupId>io.confluent.logging</groupId>
             <artifactId>common-logging</artifactId>
             <version>${confluent.version}</version>
         </dependency>
--- a/log4j2-extensions/pom.xml
+++ b/log4j2-extensions/pom.xml
@@ -31,6 +31,7 @@
     </properties>
 
     <artifactId>confluent-log4j2-extensions</artifactId>
+    <groupId>io.confluent.logging</groupId>
     <packaging>jar</packaging>
 
     <dependencies>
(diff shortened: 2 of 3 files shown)


log4j-extensions/pom.xml
```diff
@@ -31,6 +31,7 @@
     </properties>
 
     <artifactId>confluent-log4j-extensions</artifactId>
+    <groupId>io.confluent.logging</groupId>
```
ewencp: this is ok as long as this was merged for 5.2.x. if it was earlier, this would only be ok in major version bump.

real e2e validation is to run full packaging against 5.2.x but this PR branch for common. that should cover smoke tests and also provide testable artifacts if you want to do any manual validation.

Also, worth pointing out that you should be able to exclude via artifact and not just group, e.g. see https://github.com/confluentinc/schema-registry/blob/master/package-schema-registry/src/assembly/package.xml#L54 in schema reg. Not sure if there is a reason for this approach other than a) no compatibility issue and b) possibility of future additions that can be caught proactively.

rodesai: Yeah, these artifacts were introduced in 5.2.

> real e2e validation is to run full packaging against 5.2.x but this PR branch for common. that should cover smoke tests and also provide testable artifacts if you want to do any manual validation.

How do I do that?

> Also, worth pointing out that you should be able to exclude via artifact and not just group, e.g. see https://github.com/confluentinc/schema-registry/blob/master/package-schema-registry/src/assembly/package.xml#L54 in schema reg. Not sure if there is a reason for this approach other than a) no compatibility issue and b) possibility of future additions that can be caught proactively.

(b) was the big reason. The convention here seems to be that stuff packaged in common uses the group io.confluent with the artifact prefix common-. So in theory sticking to that convention will avoid future pain down the road.

ewencp: To run e2e, create a packaging branch where you override the branch for common by adding a `common_BRANCH=...` setting to `settings.sh`. If the change is only in your fork, you'll also need to adjust the `COMMON_REPO` setting. From there, a PR on the packaging branch containing those changes would run a full build as usual.

rodesai:
we've kind of missed the boat for this.

PR #182, opened 2019-03-07
Move dependency-check to only run under jenkins maven profile

ewencp:


ewencp:
I would prefer to keep this in the main developer path, but atm there seems to be some issue with db locking in Jenkins jobs that access it repeatedly, e.g. ducker builds. Most developers will probably prefer this anyway since the scanning takes time and isn't important to their normal workflow. They can always still run this locally under the jenkins profile if needed.

PR #187, opened 2019-03-29
use maven to generate archive packages

xli1996:
The reason I use a new archive.xml is because maven-assembly will append <id> to the artifact if I use the same package.xml.


package/src/assembly/archive.xml
```diff
@@ -0,0 +1,48 @@
...
+  ~ limitations under the License.
+  ~-->
+
+<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2
+          http://maven.apache.org/xsd/assembly-1.1.2.xsd">
+    <id>package</id>
+    <formats>
+        <format>zip</format>
```
ewencp: Is this basically the same as `package.xml` except for the `formats`? Could we just add these to the existing `package.xml`? (I had actually forgotten / not realized that it supported generating multiple formats in one config...)

ewencp:
Oh, now that I've actually read the comment in the PR, I'm not sure I understand the issue. What is the issue with appending a new artifact? Or am I misunderstanding?

xli1996:
Actually it is a simple issue. The `<appendAssemblyId>` will remove the `formate <id>` that appends to the artifacts. e.g. `common-archive.zip` become `common.zip`. If I use this option in `package.xml`, then the original `common-package` will become `common` directory. I am not sure do we still want to keep it. It can also be easily solved by some bash scripts.

PR #188, opened 2019-04-08
deprecate common configs

norwood:


config/src/main/java/io/confluent/common/config/ConfigUtils.java
```diff
@@ -105,4 +107,20 @@ public static Properties translateDeprecated(Properties props, String[][] synony
     }
     return newProps;
   }
+
+  /**
```
ewencp: why are we adding things while deprecating?

PR #189, opened 2019-04-09
download artifacts from nexus through restful API

xli1996:
This scripts download newest artifacts(zip/tar.gz) from nexus through restful API. It searches package_name, version, nexus's username and password and stores them in config temporary. Then the scripts uses nexus restful API to get newest artifacts and downloads those artifacts to current directory. It will be called by make archive NEW_BUILD=yes. The scripts can be potentially used by all other repos, which can save lots of time during packaging.


config.py
```diff
@@ -0,0 +1,11 @@
+class Config:
```
mohnishbasha: Is this config specific to Nexus? Can we use a more specific class name that can make its more obvious?

mohnishbasha:
@xli1996 Can we add more context to the description?
Problem that we are trying to solve.
How does this help solve the problem being solved?
Any critical impact that one needs to watch out once this change goes in?

xli1996:
This scripts download newest artifacts(zip/tar.gz) from nexus through restful API. It searches `package_name`, `version`, nexus's `username` and `password` and stores them in `config` temporary. Then the scripts uses nexus restful API to get newest artifacts and downloads those artifacts to current directory. It will be called by `make archive NEW_BUILD=yes`. The scripts can be potentially used by all other repos, which can saved lots of time during packaging.

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+    with open(m2_path) as f:
+        for line in f:
+            if re.search(r"<username>(.*)</username>", line) != None and Config.username=="":
+                Config.username = re.search(r"<username>(.*)</username>", line).groups()[0]
+            if re.search(r"<password>(.*)</password>", line) != None and Config.password=="":
+                Config.password = re.search(r"<password>(.*)</password>", line).groups()[0]
+
+def get_download_path(Config):
+    latest = 0
+    url = "https://nexus.confluent.io/service/rest/v1/search/assets?q="+Config.project_name+"-package-"+Config.version+"&maven.artifactId="+Config.project_name+"-package"
```
mohnishbasha: Define base url (BASE_URL) as constant.
Prepare the rest of the url using BASE_URL.
Rename url to nexus_url to make it more obvious.
Define "-package" & "-package-" strings as constant as well.

Infact, You can use a class to define this.
Class NexusURL:
   def BASE_URL
   def PACKAGE_LITERAL
   
    # use a function to get the prepared url
   def getNexusURL(project_name, project_version)
        return preparedNexusURL

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+                    Config.version = version.groups()[0]
+                    print(Config.version)
+            if Config.project_name=="":
+                project_name = re.search(r"<name>(.*?)(|-parent)</name>", line)
+                if project_name != None:
+                    Config.project_name = project_name.groups()[0]
+                    print(Config.project_name)
+
+def get_m2_settings(Config):
+    m2_path = os.path.expanduser("~/.m2/settings.xml")
```
mohnishbasha: This can be a property as part of NexusURL class where you can abstract the expansion, see below.

config.py
mohnishbasha: Rename to NexusConfig as it appears to be specific to you the download artifact from nexus functionality.

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+from os.path import expanduser
+
+import requests
+
+import config
+def get_project_name_version(Config):
+    with open("pom.xml") as f:
+        for line in f:
+            if Config.version=="":
+                version = re.search(r"<version>(.*?)(|-SNAPSHOT)</version>", line)
```
mohnishbasha: move the regex pattern to a constant

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+                if version != None:
+                    Config.version = version.groups()[0]
+                    print(Config.version)
+            if Config.project_name=="":
```
mohnishbasha: Instead of checking for empty string each time, you might want to use a initialization function to ensure your NexusConfig construct doesnt hold any value.

NexusConfig.init() -- sets all the properties to default value or empty.

config.py
mohnishbasha: Rename file from config.py to nexusconfig.py.

download_artifact.py
```diff
@@ -0,0 +1,64 @@
+import json
+import re
+import os.path
+from os.path import expanduser
+
+import requests
+
+import config
```
mohnishbasha: import nexusconfig

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+                if project_name != None:
+                    Config.project_name = project_name.groups()[0]
+                    print(Config.project_name)
+
+def get_m2_settings(Config):
+    m2_path = os.path.expanduser("~/.m2/settings.xml")
+    with open(m2_path) as f:
+        for line in f:
+            if re.search(r"<username>(.*)</username>", line) != None and Config.username=="":
+                Config.username = re.search(r"<username>(.*)</username>", line).groups()[0]
```
mohnishbasha: Throw exception if username or password is not found or set, since this will fail the download anyway.

mohnishbasha:
> This scripts download newest artifacts(zip/tar.gz) from nexus through restful API. It searches `package_name`, `version`, nexus's `username` and `password` and stores them in `config` temporary. Then the scripts uses nexus restful API to get newest artifacts and downloads those artifacts to current directory. It will be called by `make archive NEW_BUILD=yes`. The scripts can be potentially used by all other repos, which can save lots of time during packaging.

Please add this to the description of the pr.

config.py
mohnishbasha: why do these scripts need to be in common? can these live in common-tools?
You might also want to move these scripts under a package like confluent.tools or confluent.builds

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+def get_download_path(Config):
+    latest = 0
+    url = "https://nexus.confluent.io/service/rest/v1/search/assets?q="+Config.project_name+"-package-"+Config.version+"&maven.artifactId="+Config.project_name+"-package"
+    r = requests.get(url,auth=(Config.username,Config.password))
+    artifacts = r.json().get('items')
+    for artifact in artifacts:
+        path = artifact.get('path')
+        path_split = path.split("-"+Config.version+"-",1)
+        if len(path_split) > 1:
+            match = re.match(r"\d{8}.\d{6}-(\d*)-.*(.zip|.tar.gz)$",path_split[1])
```
mohnishbasha: define a constant for the regex pattern, add documentation as to what this intends to do.

config.py
```diff
@@ -0,0 +1,11 @@
...
+    project_name = ""
+    version = ""
+    username = ""
+    password = ""
+    download_path = {   ".zip":"",
```
mohnishbasha: keys can be simply "zip" and "tar"

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+            if match is not None:
+                if int(match.groups()[0]) >= latest:
+                    latest = int(match.groups()[0])
+                    Config.download_path[match.groups()[1]] = artifact.get('downloadUrl')
+    print(Config.download_path)
+
+def download(Config):
+    for formate,download_path in Config.download_path.items():
```
mohnishbasha: by 'formate', do you mean file_type?

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+        r = requests.get(download_path,auth=(Config.username,Config.password))
+        package_name = "confluent-"+Config.project_name+"-"+Config.version+"-SNAPSHOT"+formate
+        package = open(package_name,'wb')
+        package.write(r.content)
+        package.close()
+
+def main():
+    cg = config.TestConfig
+    get_project_name_version(cg)
```
mohnishbasha: This is not a good class/ behavior modeling. Your intention should be to define one high level behavior (lets call it, class DownloadArtifacts) that will wrap get_project_name_version, get_m2_settings, get_download_path and finally download (all of which should be private behavior since we don't want to expose them for others to call as they are incomplete by themselves).

class DownloadArtifacts:
          RepoType: "nexus" // in future we can then enhance this to handle downloads from artifactory as well.
          BaseRepoURL: "base url'
         ..... // other properties.
         init() // to initialize defaults and configs. // this can be done in the constructor as well.
         _ get_project_name_version
        _get_m2_settings
        _ get_download_path
       _downloadArtifact
      // use exceptions to handle corner cases as you validate the values from pom.xml and settings.xml.

      getArtifact() // to call all of above.

In main:
You will just initialize:
dA = DownloadArtifacts
dA.getArtifact()

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+    get_m2_settings(cg)
+    get_download_path(cg)
+    download(cg)
+
+if __name__ == "__main__":
```
mohnishbasha: Please add test cases for each of the functionality.

config.py
mohnishbasha: Remove extra space between { and "

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+import os.path
+from os.path import expanduser
+
+import requests
+
+import config
+def get_project_name_version(Config):
+    with open("pom.xml") as f:
+        for line in f:
+            if Config.version=="":
```
mohnishbasha: insert space before and after ==

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+        for line in f:
+            if re.search(r"<username>(.*)</username>", line) != None and Config.username=="":
+                Config.username = re.search(r"<username>(.*)</username>", line).groups()[0]
+            if re.search(r"<password>(.*)</password>", line) != None and Config.password=="":
+                Config.password = re.search(r"<password>(.*)</password>", line).groups()[0]
+
+def get_download_path(Config):
+    latest = 0
+    url = "https://nexus.confluent.io/service/rest/v1/search/assets?q="+Config.project_name+"-package-"+Config.version+"&maven.artifactId="+Config.project_name+"-package"
+    r = requests.get(url,auth=(Config.username,Config.password))
```
mohnishbasha: add space between , and auth.
add space between ,  and Config.password

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+    artifacts = r.json().get('items')
+    for artifact in artifacts:
+        path = artifact.get('path')
+        path_split = path.split("-"+Config.version+"-",1)
```
mohnishbasha: add space between , and 1

download_artifact.py
mohnishbasha: space between , and path_split

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+        if len(path_split) > 1:
+            match = re.match(r"\d{8}.\d{6}-(\d*)-.*(.zip|.tar.gz)$",path_split[1])
+            if match is not None:
+                if int(match.groups()[0]) >= latest:
+                    latest = int(match.groups()[0])
+                    Config.download_path[match.groups()[1]] = artifact.get('downloadUrl')
+    print(Config.download_path)
```
mohnishbasha: Add a log message for better clarity.

download_artifact.py
```diff
@@ -0,0 +1,64 @@
...
+
+def download(Config):
+    for formate,download_path in Config.download_path.items():
+        r = requests.get(download_path,auth=(Config.username,Config.password))
+        package_name = "confluent-"+Config.project_name+"-"+Config.version+"-SNAPSHOT"+formate
+        package = open(package_name,'wb')
+        package.write(r.content)
+        package.close()
```
mohnishbasha: How do you intend to handle an exception if request doesn't complete or if the file operation runs into a issue?

xli1996:
Move the script to #190, close this PR.

PR #190, opened 2019-04-09
add PULL_ARTIFACTS option

xli1996:
I move download script #189 to here to make things more clearly.

Problem: We build everything from scratch during packaging while we have already published artifacts. 

Solution: Download those artifacts directly during packaging. For this PR #190, it only reused newest `zip/tar.gz` artifacts for `common` repo. This is a POC work.

Structure:
* py/
  * confluent/
    * build/
        * download_artifacts/
            * download_artifacts.py
            * global_config.py
            * nexusURL.py
            * constants.py
            * \_\_init\_\_.py
  * test/
    * test_DownloadArtifacts.py
        * data/
            * settings.xml
            * nexus.json
            * pom.xml

The script download artifacts from nexus with option `PULL_ARTIFACTS`. I also clean a little bit for the `Makefile` to get rid of lots of `DEFAULT_XYZ` and `apply_patch`.
The script is consisted by
* `download_artifacts`: download the artifacts.
* `nexusURL`: get the right nexus download path and store it in config.
* `global_config`: all the config.
* `test_DownloadArtifacts`: unit test for high level functions, use `pytest` to run testcases.
* `data/`: test data

Since this script is only for `common` repo and other repos still rely on local maven artifacts during packaging, so I still keep `mvn install`. The `PULL_ARTIFACTS` will download artifacts and build local artifacts from scratch.

For time metric, it take around `27.045s` to run whole `mvn install`. And it take around `2.5s` to download artifacts for the first time.


Makefile
```diff
@@ -55,31 +48,35 @@ SYSCONFDIR:=$(DEFAULT_SYSCONFDIR)
...
+build:
+ifeq ($(NEW_BUILD),yes)
+	pip3 install requests
+	python3 download_artifact
+endif
 ifeq ($(SKIP_TESTS),yes)
 	mvn -DskipTests=true install
 else
-	mvn install
+		mvn install
```
mohnishbasha: fix indentation?

Makefile
```diff
@@ -17,18 +17,11 @@ export PACKAGE_NAME=$(PACKAGE_TITLE)-$(VERSION)
...
 DEFAULT_SYSCONFDIR=PREFIX/etc/$(PACKAGE_TITLE)
 DEFAULT_SKIP_TESTS=no
-
-
-# Whether we should apply patches. This only makes sense for alternate packaging
-# systems that know how to apply patches themselves, e.g. Debian.
-ifndef APPLY_PATCHES
-APPLY_PATCHES=$(DEFAULT_APPLY_PATCHES)
-endif
+DEFAULT_NEW_BUILD=no
```
mohnishbasha: DEFAULT_NEW_BUILD - its not clear what is NEW_BUILD and when this needs to be set.

Could you add a comment as to why this is being set?
When set to yes, what should be the behavior?

Looks like when set to yes, it downloads from nexus.
Literal DEFAULT_NEW_BUILD doesn't make it obvious, may be you can use a better name here.
How about DEFAULT_PULL_ARTIFACTS ?

Makefile
```diff
@@ -55,31 +48,35 @@ SYSCONFDIR:=$(DEFAULT_SYSCONFDIR)
...
+ifndef NEW_BUILD
+NEW_BUILD=$(DEFAULT_NEW_BUILD)
+endif
+
 export VERSION
 export DESTDIR
 export PREFIX
 export SYSCONFDIR
 export SKIP_TESTS
+export NEW_BUILD
```
mohnishbasha: PULL_ARTIFACTS

Makefile
```diff
@@ -55,31 +48,35 @@ SYSCONFDIR:=$(DEFAULT_SYSCONFDIR)
 endif
 SYSCONFDIR:=$(subst PREFIX,$(PREFIX),$(SYSCONFDIR))
 
-export APPLY_PATCHES
+ifndef NEW_BUILD
+NEW_BUILD=$(DEFAULT_NEW_BUILD)
```
mohnishbasha: PULL_ARTIFACTS=$(DEFAULT_PULL_ARTIFACTS)
if enabled, this will pull from nexus/ (in future artifactory), else will locally build?

mohnishbasha:
Is there a reason, why this is being merged into archive branch?

ewencp:
@mohnishbasha `common` is still on `archive`/`rpm`/`deb` branches. @xiangxin72 We probably would want to go through and collapse/standardize as we do all these updates. Doesn't have to happen before this PR, but since we'll be modifying all the repos anyway, it's a good opportunity to finally follow through on it.

Makefile
```diff
@@ -55,31 +48,35 @@ SYSCONFDIR:=$(DEFAULT_SYSCONFDIR)
...
-apply-patches: $(wildcard patches/*)
-ifeq ($(APPLY_PATCHES),yes)
-	git reset --hard HEAD
-	cat patches/series | xargs -iPATCH bash -c 'patch -p1 < patches/PATCH'
 endif
 
-build: apply-patches
+build:
+ifeq ($(NEW_BUILD),yes)
+	pip3 install requests
```
ewencp: we should be fine having this installed generally, so this would be better moved to the code that generates the AMIs https://github.com/confluentinc/confluent-amis so it is available in all builds. Otherwise all the component repos will have to duplicate this code to install `requests`

Makefile
ewencp: I don't remember why these ended up written this way (probably because way back when I made a bad choice) but we don't need to separately define the `DEFAULT_XYZ` variable and then use it to override the default, we can just inline these into the line that sets the `XYZ` variable.

I agree on naming -- could be clearer. `PULL_ARTIFACTS` or something around "prebuilt" maybe.

Makefile
```diff
@@ -55,31 +48,35 @@ SYSCONFDIR:=$(DEFAULT_SYSCONFDIR)
...
+	python3 download_artifact
```
ewencp: i think we're missing the `download_artifact` script?

xli1996: I actually put the scripts in the master branch because I think it can be reused later by other branches. However, just for archive, I should move the scripts on this branch.

ewencp: ah, i see, link relevant PR? anyone looking at history will appreciate the connection since it's hard to reason about PRs that are dependent on each other without links.

additionally, in cases like this it is helpful to link the other PR since the other PR needs to be merged *before* this one is merged. understanding merge dependency order also helps with PR review and allows us to make sure we merge in the right order

xli1996:
I move download script #189 to here to make things more clearly.

The script download artifacts from nexus with option `PULL_ARTIFACTS`. I also clean a little bit for the `Makefile` to get rid of lots of `DEFAULT_XYZ` and `apply_patch`.
The script is consisted by
`download_artifacts`: get the right package name&version and download the package.
`nexusURL`: get the right nexus download path and store it in config.
`global_config`: all the config.

I resolved lots of comments in #189, but there are fews I haven't solved like the testcases. Since the `archive` branch doesn't go through Jenkins currently. And during the packaging, it may not support `tox` and `pytest`? Someone can help me with it. I will add more pytest later after when we apply it to other repos. However, for now those scripts are simple tests specifically towards `common` repo. I want to see whether it could cause any potential problem during the packaging process.

mohnishbasha:
Fix PR title, current title is no longer relevant as it refers to NEW_BUILD.

Makefile
```diff
@@ -17,65 +17,48 @@ export PACKAGE_NAME=$(PACKAGE_TITLE)-$(VERSION)
...
-apply-patches: $(wildcard patches/*)
-ifeq ($(APPLY_PATCHES),yes)
-	git reset --hard HEAD
-	cat patches/series | xargs -iPATCH bash -c 'patch -p1 < patches/PATCH'
 endif
 
-build: apply-patches
+build:
+ifeq ($(PULL_ARTIFACTS),no)
+	python3 ./download_artifacts/download_artifacts.py
```
mohnishbasha: if this is more like a one of script, better to move it under a common/bin directory instead of common/download_artifacts/ directory.

if not, better to keep it under a python package, common/py/confluent/build/download_artifacts.py

download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,65 @@
...
+from global_config import DAConfig
+from nexusURL import NexusURL
+
+VERSION_MATCH = r"<version>(.*?)(|-SNAPSHOT)</version>"
+NAME_MATCH = r"<name>(.*?)(|-parent)</name>"
+SNAPSHOT_LITERAL = "-SNAPSHOT"
+CONFLUENT_LITERAL = "confluent-"
+POM_PATH = "pom.xml"
+
+class DownloadArtifacts:
```
mohnishbasha: Add documentation.
Refer here: 
https://github.com/confluentinc/release-tools/blob/master/confluent/release/constants.py
https://github.com/confluentinc/release-tools/blob/bcd5dd8bd6aa3d335a8b689948f212da7b068c57/confluent/release/time_based.py#L28

download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,65 @@
+import json
+import re
+import sys
+
+import requests
+
+from global_config import DAConfig
+from nexusURL import NexusURL
+
+VERSION_MATCH = r"<version>(.*?)(|-SNAPSHOT)</version>"
```
mohnishbasha: VERSION_REGEX
NAME_REGEX

For Literals, use ENUM:
Refer here: https://github.com/confluentinc/release-tools/blob/master/confluent/release/constants.py

You might not want to expose these REGEX and LITERAL constants as public, instead encapsulate them within the class.

download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,65 @@
...
+NAME_MATCH = r"<name>(.*?)(|-parent)</name>"
+SNAPSHOT_LITERAL = "-SNAPSHOT"
+CONFLUENT_LITERAL = "confluent-"
+POM_PATH = "pom.xml"
+
+class DownloadArtifacts:
+    def __init__(self, cg):
+        self.cg = cg
+
+    def get_project_name_version(self):
```
mohnishbasha: Add documentation.

download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,65 @@
...
+                    if self.cg.version == "":
+                        version = re.search(VERSION_MATCH, line)
+                        if version is not None:
+                            self.cg.version = version.groups()[0]
+                    if self.cg.project_name == "":
+                        project_name = re.search(NAME_MATCH, line)
+                        if project_name is not None:
+                            self.cg.project_name = project_name.groups()[0]
+        except IOError as e:
+            print(e)
```
mohnishbasha: You might want to elaborate the exception with a message.

download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+    def download(self):
```
mohnishbasha: rename to download_artifact to make it more obvious to the reader.

download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,65 @@
...
+            except requests.exceptions.RequestException as e:
+                print(e)
+                sys.exit(1)
+            package_name = CONFLUENT_LITERAL + self.cg.project_name + "-" +\
+                self.cg.version + SNAPSHOT_LITERAL + file_type
+            try:
+                package = open(package_name, 'wb')
+                package.write(r.content)
+                package.close()
+            except IOError as e:
```
mohnishbasha: instead use one try except block:
try:
.... 
except RequestException:
except IOError

download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+    def download(self):
+        for file_type, download_path in self.cg.download_path.items():
+            try:
+                r = requests.get(
+                        download_path, auth=(self.cg.username, self.cg.password))
+            except requests.exceptions.RequestException as e:
+                print(e)
+                sys.exit(1)
```
mohnishbasha: Remove sys.exit. 
Handle the exception, print the error message and stack trace.
If you need to, throw the exception again.

download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def main():
+    cg = DAConfig
+    da = DownloadArtifacts(cg)
+    da.get_project_name_version()
+    nurl = NexusURL(cg)
+    nurl.get_m2_settings()
+    nurl.get_download_path()
+    da.download()
```
mohnishbasha: da = DownloadArtifacts()
da.download_artifacts()

These are the only 2 lines that should matter here in main.
Everything else should happen within DownloadArtifacts()

download_artifacts/global_config.py
```diff
@@ -0,0 +1,9 @@
+class DAConfig:
+    project_name = ""
+    version = ""
+    username = ""
+    password = ""
+    download_path = {".zip": "",
```
mohnishbasha: Better use a child object:
zip: {
file_ext: ".zip",
file_url: ""
}
tar: {
...
...
}

Makefile
```diff
@@ -17,65 +17,48 @@ export PACKAGE_NAME=$(PACKAGE_TITLE)-$(VERSION)
...
-# Whether we should run tests during the build.
-ifndef SKIP_TESTS
-SKIP_TESTS=$(DEFAULT_SKIP_TESTS)
-endif
 
 # Install directories
-ifndef DESTDIR
-DESTDIR=$(DEFAULT_DESTDIR)
-endif
+DESTDIR=$(CURDIR)/BUILD/
```
mohnishbasha: You might want to keep ifndef

download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,47 @@
+import os
+import re
+import sys
+
+import requests
+
+BASE_URL = "https://nexus.confluent.io/service/rest/v1/search/assets?q="
```
mohnishbasha: Encapsulate these within NexusURL.
For CONTSTANTS  you can also use ENUM.

download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,47 @@
...
+USERNAME_MATCH = r"<username>(.*)</username>"
+PASSWORD_MATCH = r"<password>(.*)</password>"
+FILENAME_MATCH = r"\d{8}.\d{6}-(\d*)-.*(.zip|.tar.gz)$"
+
+
+class NexusURL:
+    def __init__(self, cg):
+        self.cg = cg
+
+    def get_m2_settings(self):
```
mohnishbasha: rename to get_maven_settings

download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,47 @@
...
+        m2_path = os.path.expanduser(M2_PATH)
+        with open(m2_path) as f:
+            for line in f:
+                if re.search(USERNAME_MATCH, line) is not None and self.cg.username == "":
+                    self.cg.username = re.search(USERNAME_MATCH, line).groups()[0]
+                if re.search(PASSWORD_MATCH, line) is not None and self.cg.password == "":
+                    self.cg.password = re.search(PASSWORD_MATCH, line).groups()[0]
+
+    def get_download_path(self):
```
mohnishbasha: rename get_artifact_uri

download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,47 @@
...
+        latest = 0
+        url = BASE_URL + self.cg.project_name + PACKAGE_LITERAL + "-" + self.cg.version + ARTIFACTID + self.cg.project_name + PACKAGE_LITERAL
+        try:
+            r = requests.get(url, auth=(self.cg.username, self.cg.password))
+        except requests.exceptions.RequestException as e:
+            print(e)
+            sys.exit(1)
```
mohnishbasha: not a good idea to keep sys.exit as it will quit the entire process.

download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,47 @@
...
+        for artifact in artifacts:
+            path = artifact.get('path')
+            path_split = path.split("-"+self.cg.version+"-", 1)
+            if len(path_split) > 1:
+                match = re.match(FILENAME_MATCH, path_split[1])
+                if match is not None:
+                    if int(match.groups()[0]) >= latest:
+                        latest = int(match.groups()[0])
+                        self.cg.download_path[match.groups()[1]] = artifact.get('downloadUrl')
+        print(self.cg.download_path)
```
mohnishbasha: Add a message to make it more clear.
"Artifact URL: " ...

mohnishbasha:
For each of the high level function, add a test case.

py/test/test_DownloadArtifacts.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@mock.patch('builtins.open', new_callable=mock_open, create=True)
+def test_get_maven_settings(mock_open):
+    mock_open.return_value.__enter__ = mock_open
+    mock_open.return_value.__iter__ = mock.Mock(
+        return_value=iter(test_maven_settings))
+    da = DownloadArtifacts()
+    da.get_maven_settings()
+    assert da.cg.username == "xiang username"
```
mohnishbasha: lets not use concrete names as such. use a some generic temp name if needed.

py/test/test_DownloadArtifacts.py
```diff
@@ -0,0 +1,73 @@
...
+
+@patch.object(Session, 'get')
+def test_get_artifact_uri(mock_session):
+    mock_session.return_value = MockResponse(nexus_result, 200)
+    da = DownloadArtifacts()
+    da.get_artifact_uri()
+    assert da.cg.artifacts == {
+                                'zip': {
+                                    'file_ext': '.zip',
+                                    'file_url': 'https://nexus.confluent.io/repository/maven-snapshots/io/confluent/common-package/5.3.0-SNAPSHOT/common-package-5.3.0-20190413.162837-58-package.zip'
```
mohnishbasha: This is break if nexus run into issues or the artifact gets moved from the path or deleted.
Use mock or prepare the test data locally.

py/test/data/settings.xml
```diff
@@ -0,0 +1,48 @@
...
+ 
+ <servers>
+   <server>
+     <id>confluent-nexus-central</id>
+     <username>xiang username</username>
+     <password>xiang password</password>
+   </server>
+   <server>
+     <id>confluent-maven-remote</id>
+     <username>xiang username</username>
```
mohnishbasha: user a generic username .. "nexus_user"

Makefile
```diff
@@ -47,49 +30,60 @@ endif
...
 endif
 
 ifndef SYSCONFDIR
-SYSCONFDIR:=$(DEFAULT_SYSCONFDIR)
+SYSCONFDIR=PREFIX/etc/$(PACKAGE_TITLE)
+endif
+
+# Whether we pull artifacts from nexus/artifactory
+ifndef PULL_ARTIFACTS
+PULL_ARTIFACTS=no
```
mohnishbasha: You can set default value to shell variables as in like here:
https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash

if env variable is available use it else use default.

py/confluent/build/download_artifacts/constants.py
```diff
@@ -0,0 +1,15 @@
+from enum import Enum
```
mohnishbasha: It doesn't look like you are using Enums.

py/confluent/build/download_artifacts/constants.py
```diff
@@ -0,0 +1,15 @@
...
+
+BASE_URL = "https://nexus.confluent.io/service/rest/v1/search/assets?q="
```
mohnishbasha: These are simply literal constants.

py/confluent/build/download_artifacts/constants.py
```diff
@@ -0,0 +1,15 @@
...
+ARTIFACTID = "&maven.artifactId="
+USERNAME_REGEX = r"<username>(.*)</username>"
+PASSWORD_REGEX = r"<password>(.*)</password>"
+FILENAME_REGEX = r"\d{8}.\d{6}-(\d*)-.*(.zip|.tar.gz)$"
+
+VERSION_REGEX = r"<version>(.*?)(|-SNAPSHOT)</version>"
+NAME_REGEX = r"<name>(.*?)(|-parent)</name>"
+SNAPSHOT_LITERAL = "-SNAPSHOT"
+CONFLUENT_LITERAL = "confluent-"
+POM = "pom.xml"
```
mohnishbasha: Refer this to see how to define enums:
https://github.com/confluentinc/release-tools/blob/30d1139fb626789dfd3cb448ef7879b2fc457343/confluent/release/constants.py#L13

py/confluent/build/download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,77 @@
+import json
+import re
+
+import requests
+
+from global_config import DownloadArtifactsConfig
+from nexusURL import NexusURL
+from constants import VERSION_REGEX, NAME_REGEX, SNAPSHOT_LITERAL, CONFLUENT_LITERAL, POM
```
mohnishbasha: You can address this with Enum, one you define one.

py/confluent/build/download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,77 @@
+import json
+import re
+
+import requests
+
+from global_config import DownloadArtifactsConfig
+from nexusURL import NexusURL
```
mohnishbasha: change name to nexus_url  (python guideline).
NexusURL (this is java style)

py/confluent/build/download_artifacts/constants.py
```diff
@@ -0,0 +1,15 @@
+from enum import Enum
+
+BASE_URL = "https://nexus.confluent.io/service/rest/v1/search/assets?q="
+M2 = "~/.m2/settings.xml"
```
mohnishbasha: MVN_SETTINGS

py/confluent/build/download_artifacts/constants.py
mohnishbasha: Some good info on how to style python using pep8
https://www.python.org/dev/peps/pep-0008/#constants

py/confluent/build/download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,77 @@
...
+from constants import VERSION_REGEX, NAME_REGEX, SNAPSHOT_LITERAL, CONFLUENT_LITERAL, POM
+import time
+
+
+class DownloadArtifacts:
```
mohnishbasha: Add documentation

py/confluent/build/download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,77 @@
+import json
+import re
+
+import requests
+
```
mohnishbasha: remove space.

native imports

3rd party imports

confluent imports

(ordered alphabetically .. )

py/confluent/build/download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,77 @@
...
+            try:
+                r = requests.get(
+                        artifact.get('file_url'), auth=(self.cg.username, self.cg.password))
+                package_name = CONFLUENT_LITERAL + self.cg.project_name + "-" +\
+                    self.cg.version + SNAPSHOT_LITERAL + artifact.get('file_ext')
+                package = open(package_name, 'wb')
+                package.write(r.content)
+                package.close()
+            except requests.exceptions.RequestException as e:
+                print(e)
```
mohnishbasha: you can add a verbose message for better context along with printing the exception

py/confluent/build/download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,77 @@
...
+
+from global_config import DownloadArtifactsConfig
+from nexusURL import NexusURL
+from constants import VERSION_REGEX, NAME_REGEX, SNAPSHOT_LITERAL, CONFLUENT_LITERAL, POM
+import time
+
+
+class DownloadArtifacts:
+    def __init__(self):
+        cg = DownloadArtifactsConfig
```
mohnishbasha: use clear names to make it obvious to the read.
config
nexus_url

py/confluent/build/download_artifacts/download_artifacts.py
```diff
@@ -0,0 +1,77 @@
...
+        self.get_artifact_uri()
+        self.download_artifact()
+        return
+
+
+def main():
+    start_time = time.time()
+    da = DownloadArtifacts()
+    da.download()
+    print("--- %s seconds ---" % (time.time() - start_time))
```
mohnishbasha: Add a clear message .. simple printing the time in seconds doesnt provide any context.

py/confluent/build/download_artifacts/global_config.py
```diff
@@ -0,0 +1,16 @@
+class DownloadArtifactsConfig:
```
mohnishbasha: Add documentation, what is the purpose of this struct. What do you intend to do with this.

py/confluent/build/download_artifacts/global_config.py
```diff
@@ -0,0 +1,16 @@
...
+    artifacts = {
+                    "zip":{
+                        "file_ext":".zip",
+                        "file_url":""
+                    },
+                     "tar.gz":{
+                         "file_ext":".tar.gz",
+                         "file_url":""
+                     }
+                    }
```
mohnishbasha: looks like the json is not aligned right.
Use pep8 to validate ur python code.

py/confluent/build/download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,48 @@
+import os
+import re
+
+import requests
+
+from constants import BASE_URL, M2, PACKAGE_LITERAL, ARTIFACTID, USERNAME_REGEX, PASSWORD_REGEX, FILENAME_REGEX
```
mohnishbasha: Use enums.
Follow import order.

Refer to import order here for example:
https://github.com/confluentinc/release-tools/blob/master/confluent/release/projects_editor.py

py/confluent/build/download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class NexusURL:
```
mohnishbasha: Documentation.

py/confluent/build/download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, cg):
+        self.cg = cg
```
mohnishbasha: nexus_config

py/confluent/build/download_artifacts/nexusURL.py
```diff
@@ -0,0 +1,48 @@
...
+                    if re.search(PASSWORD_REGEX, line) is not None and self.cg.password == "":
+                        self.cg.password = re.search(PASSWORD_REGEX, line).groups()[0]
+        except IOError as e:
+            print(e)
+        return
+
+    def get_artifact_uri(self):
+        '''get latest artifact url'''
+        latest = 0
+        url = BASE_URL + self.cg.project_name + PACKAGE_LITERAL + "-" + self.cg.version + ARTIFACTID + self.cg.project_name + PACKAGE_LITERAL
```
mohnishbasha: use enums.

py/test/test_DownloadArtifacts.py
```diff
@@ -0,0 +1,73 @@
+import os
+import sys
+import json
+import unittest.mock as mock
+from unittest.mock import patch, mock_open
+import gzip
+
+from requests import Session
+
```
mohnishbasha: set import order

py/test/test_DownloadArtifacts.py
```diff
@@ -0,0 +1,73 @@
...
+path = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(path+'/../confluent/build/download_artifacts')
+
+from download_artifacts import DownloadArtifacts
```
mohnishbasha: make it part of imports.

py/test/test_DownloadArtifacts.py
```diff
@@ -0,0 +1,73 @@
+import os
+import sys
+import json
+import unittest.mock as mock
+from unittest.mock import patch, mock_open
+import gzip
+
+from requests import Session
+
+path = os.path.dirname(os.path.abspath(__file__))
```
mohnishbasha: move this below imports.

py/test/test_DownloadArtifacts.py
```diff
@@ -0,0 +1,73 @@
...
+test_maven_settings = []
+with open(path+'/data/settings.xml') as settings:
+    for line in settings:
+        test_maven_settings.append(line)
+
+nexus_result = json.loads(gzip.open(path+'/data/nexus.json.gz').read())
+
+
+@mock.patch('builtins.open', new_callable=mock_open, create=True)
+def test_get_project_name_version(mock_open):
```
mohnishbasha: add documentation as to what do you intend to do with this test.

py/test/test_DownloadArtifacts.py
xli1996: All data mock return is stored locally. `mock_session` should only return json data that I store in `data/nexus.json`, so it should be fine.

py/test/test_DownloadArtifacts.py
xli1996: I need to set sys path before import parent directory, is there anyway I can do it? By set sys path through cli?

Makefile
xli1996: Currently, I don't want to modify too much for `Makefile`, incase it breaks other stuff during packaging that I am not familiar with. However, I think we do need to simplify them later.

py/confluent/build/download_artifacts/download_artifacts.py
xli1996: Blank line is required between different types of imports? I will clean them alphabetically.

mohnishbasha: thats the practice we are following. refer to other similar imports in release-tools.

PR #191, opened 2019-04-14
Attaching sources jar

NersesAM:
Sources jar are missing from maven repo which makes it hard to debug in IDE when sources for given version can be downloaded with one click, whereas now one has to go clone github repo and attach sources manually.
example:
http://packages.confluent.io/maven/io/confluent/kafka-avro-serializer/5.2.1/kafka-avro-serializer-5.2.1.jar <- works great
http://packages.confluent.io/maven/io/confluent/kafka-avro-serializer/5.2.1/kafka-avro-serializer-5.2.1-sources.jar <- doesn't 😞

PR #192, opened 2019-04-14
Add maven source plugin to attach sources

worldtiki:
It's painful to troubleshoot issues when things go wrong without having access to the sources.
I've been cloning the different projects and attaching them as sources, but it would be great if the sources jars were also published.


worldtiki:
Just noticed the exact same issue was raised a few minutes ago in #191 :D

Closing this one

PR #194, opened 2019-05-06
add newbuild

xli1996:
**Goal**: Remove download scripts form common repo.
**Changes**: Add a new target called `newbuild` which can be trigged by `PULL_ARTIFACTS`.

**Explain**:
The `newbuild` will clone and execute the download artifacts scripts and remove it.
Also, I remove the `./create_archive.sh` from `newbuild` cause I think it is not needed if artifacts are directly downloaded from nexus.


Makefile
```diff
@@ -65,18 +65,24 @@ ifeq ($(PULL_ARTIFACTS),no)
...
-
 ifeq ($(SKIP_TESTS),yes)
 	mvn -DskipTests=true install
 else
 	mvn install
 endif
 
-install: build
+newbuild:
+	git clone git@github.com:confluentinc/download-artifacts.git
```
ewencp: 1. do these need their own dedicated repo or could we just put them, e.g., release-tools
2. do we need versioning on these to maintain reproducible  builds? for example, what if on a future release branch we need to  change behavior of the download scripts, but we still need to be able to release older branches. The downside however, is if these versions get scattered through all the repos, it's a pain to update. (An alternative if we do keep the scripts in  their own repo is to manage versions along with product version  branches and tags, i.e. 5.1.x, 5.2.x, etc, and branch them with tbr like we do for the product repos).

Makefile
```diff
@@ -65,18 +65,24 @@ ifeq ($(PULL_ARTIFACTS),no)
...
+	python3 -u ./download-artifacts/py/confluent/build/download_artifacts/download_artifacts.py
+	rm -rf ./download-artifacts
+
+install:
+ifeq ($(PULL_ARTIFACTS),yes) 
+	make newbuild
```
ewencp: you can make the dependency conditional rather than invoking make indirectly, e.g. similar to like they use here https://stackoverflow.com/a/38353274/841827

Makefile
```diff
@@ -65,18 +65,24 @@ ifeq ($(PULL_ARTIFACTS),no)
...
+else 
+	make build
```
ewencp: i'm not sure it's actually safe to get rid of `create_archive.sh` fully here for the new build  path.  You can move the commands into the Makefeile (e.g. as has ben done for many of the  services, such as https://github.com/confluentinc/schema-registry/blob/master/debian/Makefile#L69), but that file sometimes does some rearrangement of artifacts (e.g. dealing with  PREFIX vs not), uses `install` to set user/group/permissions properly, filters some extra  stuff (looks like `README.rpm` commonly, though we should just filter that out  earlier if we  want).

Makefile
xli1996: One potential solution I am thinking is that keep my script independent from release-tools and use same version branch as CP to control behavior.

PR #198, opened 2019-05-23
Update jackson to 2.9.9

cmccabe:


ijuma:
@cmccabe I think you should backport this to 5.3.x since there's a CVE (even if medium).

PR #206, opened 2019-06-14
Ability to define alternative configurations

vvcephei:
Add the ability to define exclusive alternatives in configurations (such as "A or B must be specified").
* alternatives are associated by virtue of an "alternative group name", which is purely used for grouping, not visible in the config definition.
* at most one of the alternatives can have a default value. If present, this becomes the default value of the default config for the group. Otherwise, it's required that exactly one config in the group be specified.

### DIFF
--- a/config/src/main/java/io/confluent/common/config/AbstractConfig.java
+++ b/config/src/main/java/io/confluent/common/config/AbstractConfig.java
@@ -80,6 +80,10 @@ public class AbstractConfig {
     logAll();
   }
 
+  protected boolean contains(String key) {
+    return values.containsKey(key);
+  }
+
   protected Object get(String key) {
     if (!values.containsKey(key)) {
       throw new ConfigException(String.format("Unknown configuration '%s'", key));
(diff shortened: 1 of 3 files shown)


vvcephei:
[clabot:check]

vvcephei:
@ewencp , do you mind taking a look at this, since according to the git-praise, you have some contact with config parsing?

ewencp:
I guess I appreciate the term git-praise instead of git-blame :)

Note that we'd want to upstream this since we are trying to get rid of this copy of the AK configdef stuff.

In Connect, we added additional validation step to handle stuff like this. Obviously that doesn't generalize, but if we were to do something like this, we should make sure we have a clear set of use cases and make sure it is generalized enough. I'm unclear whether we really need this or can just use a single config with sufficiently flexible config values to express alternatives...

vvcephei:
:) Yeah, "blame" isn't appropriate, especially when you're asking someone for a review!

Thanks for the feedback, here and downstream. You presented an idea for the dependent PR that wouldn't require this feature, so maybe I can just close this one.

PR #208, opened 2019-06-20
Add the buildnumber plugin and use it to add the commit sha to jars

rodesai:
This patch adds the buildnumber plugin to the common parent pom, and
uses it to add the commit sha to jars. The buildnumber plugin populates
the attribute ${buildNumber} with the git sha, which is then picked up
by the maven archiver and added to the jar manifest.

Build number plugin wiki:
https://www.mojohaus.org/buildnumber-maven-plugin/

Usage:
https://www.mojohaus.org/buildnumber-maven-plugin/usage.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,7 @@
         <maven-surefire-plugin.version>2.21.0</maven-surefire-plugin.version>
         <maven-site-plugin.version>3.7.1</maven-site-plugin.version>
         <maven-project-info-reports-plugin.version>3.0.0</maven-project-info-reports-plugin.version>
+        <buildnumber-plugin.version>1.4</buildnumber-plugin.version>
         <mockito.version>2.23.0</mockito.version>
         <mockito-all.version>1.10.19</mockito-all.version>
         <os-maven-plugin.version>1.5.0.Final</os-maven-plugin.version>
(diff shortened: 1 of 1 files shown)


mohnishbasha:
Could you add the exact line item format that is added to the jar manifest.

rodesai:
@mohnishbasha 

```
$ jar xf ksql-rest-app-5.4.0-SNAPSHOT.jar
$ cat META-INF/MANIFEST.MF
Manifest-Version: 1.0
Built-By: rohan
Commit-ID: ac2e335f32cd1403db6d7eb1e97dadb16c642e9e
Created-By: Apache Maven 3.5.2
Build-Jdk: 1.8.0_151

$
```

mohnishbasha:
Can we change the attribute buildNumber to be more specific as in commitId or gitSha?

https://www.mojohaus.org/buildnumber-maven-plugin/create-mojo.html#buildNumberPropertyName
Has a way to rename the attribute.

pom.xml
```diff
@@ -530,6 +537,24 @@
                     <artifactId>maven-dependency-plugin</artifactId>
                     <version>${maven-dependency-plugin.version}</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>buildnumber-maven-plugin</artifactId>
+                    <version>1.4</version>
```
mohnishbasha: Can we move this into the version properties up above in the pom:
https://github.com/confluentinc/common/blob/3d5b2b91fe41b7f030447f2251a4b45dabe2db8b/pom.xml#L110

PR #213, opened 2019-07-12
Fix version references for licenses and assembly tasks

sirianni:
The `create-licenses-for-docker` task should use `${licenses.version}` to match the corresponding dependency.
The `make-assembly-for-docker` task should use `${confluent.version}` to match the corresponding dependency.

This is needed for projects which inherit from this POM but set a different `project.version`.  This was the original intent of 3eec8d831e9b3ae0ac25563a1162afdae4e0b5bf but has regressed for the `docker` profile.

See [this build](https://confluentinc.semaphoreci.com/jobs/eeea0cfc-c8de-45ea-9da0-65b6f745fe38#L177) for an example.


sirianni:
[clabot:check]

PR #215, opened 2019-07-31
MINOR: Added ability to set property to skip integration tests, running them by default as before

rhauch:


pom.xml
```diff
@@ -64,6 +64,12 @@
         <!-- <jackson-databind.version>${jackson.version}</jackson-databind.version> -->
         <jackson-databind.version>2.9.9.2</jackson-databind.version>
 
+        <!-- The following property defines whether integration tests should be skipped.
+        This is set to false so they are executed by default, but since Jenkins currently
+        doesn't support builds using Docker containers, some projects that use Docker for their
+        integration tests may need to set this to false in their Jenkins profile. -->
```
norwood: false -> true?

PR #217, opened 2019-08-07
MINOR: upgrade jackson-databind to 2.9.9.3

xvrl:
Some downstream components are hitting the following regression in 2.9.9.2
https://github.com/FasterXML/jackson-databind/issues/2395


xvrl:
cc @ewencp

PR #220, opened 2019-08-30
ST-2035: Adding the scala version as a property to docker builds

elismaga:
this property will be used by some of the new docker image builds so i am putting it in the common pom docker step so it doesn't have to be defined in several pom files with in the docker repos.


elismaga:
> lgtm, however if certain projects need a different version of scala, how will this get overridden?
> Should the downstream project override this in their respective pom?

yes, the downstream pom can override this.

PR #228, opened 2019-11-15
CC-7031: Bump Jackson version to match AK

C0urante:
[Jira](https://confluentinc.atlassian.net/browse/CC-7031)

https://github.com/apache/kafka/pull/7411 bumped the Jackson dependency version in AK to 2.10.0; these changes bring CP up to speed with that change.

This change has _not_ been approved as a blocker yet, and should not be merged until/unless it is.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,8 +55,8 @@
         <spotbugs.version>3.1.8</spotbugs.version>
         <spotbugs.maven.plugin.version>3.1.8</spotbugs.maven.plugin.version>
         <java.version>8</java.version>
-        <jackson.version>2.9.10</jackson.version>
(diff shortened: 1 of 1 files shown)


arvindth:
@C0urante note that I've recently tested this change. One reason to hold off on this is that some of the downstream dependencies of common, such as schema-registry, have a dependency on the latest avro 1.9.1, which in turn has a hard dependency on jackson 2.9 due to avro's dependence on some jackson 2.9.x apis that have been deleted in jackson 2.10.0.

See apache/avro#675 for the specific method calls I'm referring to. Unfortunately, these changes are targeted to the as-yet-unreleased avro 1.10.0

C0urante:
Thanks, @arvindth. I'll close this for now and pursue an alternative fix for CC-7031.

PR #235, opened 2019-12-17
MINOR: Bump Guava version to pass dependency check

C0urante:
The current Guava version in the POM fails the dependency check stage. This new version does not.

The change here should not have any adverse consequences for repose that utilize this property since any that do should already be failing to build.

We should hold off on merging this until we're out of the blocker phase for 5.4.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
         <jackson.bom.version>2.9.10.20191020</jackson.bom.version>
 
         <gson.version>2.8.5</gson.version>
-        <guava.version>24.0-jre</guava.version>
+        <guava.version>28.1-jre</guava.version>
         <jacoco-maven-plugin.version>0.8.2</jacoco-maven-plugin.version>
         <junit.version>4.12</junit.version>
         <kafka.version>2.2.2-SNAPSHOT</kafka.version>


ewencp:
@C0urante is this still needed? it looks like the builds are passing on this branch and I recall seeing some other PR(s) updating dependency check related stuff.

C0urante:
@ewencp I think it's still needed. Since Guava isn't drawn in as a dependency here it won't cause the build to fail, but for any projects that do draw it in as a dependency, the check either has to be suppressed ([which it isn't](https://github.com/confluentinc/common/blob/6fef0d74258c018cc476c4b2d2fecda1895783eb/build-tools/src/main/resources/dependency-check/suppressions.xml), currently) or the version has to be overridden. The idea here is to bump the version in the CP parent so that any child projects that add Guava down the road won't have to worry about any of that.

ewencp:
Sure, I get the basic need, but I couldn't find a case where this was actually causing a failure. Some grepping through repos shows that most places don't even seem to use the inherited value, they just specify their own (and frequently substantially older versions). Upgrading is fine, but we should also do a build against something like packaging/system tests with this change to validate as Guava does sometimes make backwards incompatible changes.

C0urante:
> most places don't even seem to use the inherited value, they just specify their own
Well yeah, you have to override the inherited value in order for builds to pass. The point of this PR is to make it possible to draw in guava without specifying a value and instead just inherit from whatever we use in common.

If we merge this PR right now it won't bump the Guava version for any current repos (unless there are ones out there that explicitly disable the dependency check for Guava) since they all use overridden versions (otherwise their builds would be failing right now).

Follow-up PRs could then be issued on a per-repo basis to ensure that those repos work with the new Guava version established here, and any new repos added (or repos that add Guava as a dependency) wouldn't have to specify their own value and could instead inherit from common.

ewencp:
I think updating the guava version is fine, I am just not sure we want to do it on older release branches unless it is needed there -- any change on older branches introduces risk to bug fix releases. But the general approach of cleaning up overrides after updating the version is 👍 . I guess if nothing is even using this variable and everything is overriding it though, guess it doesn't much matter merging this update to older branches.

C0urante:
Thanks, Ewen!

C0urante:
Pint merge was rejected with this message:

> remote: error: GH006: Protected branch update failed for refs/heads/5.3.x.
> remote: error: At least 1 approving review is required by reviewers with write access.

@ewencp do you want to take care of the pint merge?

ewencp:
Sure, done.

PR #237, opened 2020-01-03
update cve to json feed

xli1996:
xml feed is removed form 2020, https://nvd.nist.gov/vuln/data-feeds. We should switch to json feed.


dnozay:
@xli1996 i can take another look after merge conflicts are fixed, but otherwise LGTM

PR #241, opened 2020-01-10
CC-7633: ConfigDef needs to check current thread's classloader when parsing a type CLASS config

kkonstantine:
After KAFKA-8340, KAFKA-8819 were addressed in apache/kafka this ConfigDef class needs also to consider the current thread's classloader when parsing the type CLASS config type. 

This brings parsing for these types closer to the equivalent class in apache/kafka: org.apache.kafka.common.config.ConfigDef#parseType


kkonstantine:
Working on a unit test. Will update PR as soon as I figure out viability.

kkonstantine:
Added the necessary test. This is ready for review on my side.

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
```diff
@@ -275,4 +285,79 @@ private void testValidators(Type type, ConfigDef.Validator validator, Object def
...
+    Set<String> allowedKlasses = Collections.singleton(configNameDefault);
+    List<URL> pluginUrls = Collections.singletonList(Paths.get("target/classes/").toUri().toURL());
+    URLClassLoader pluginLoader = new PluginClassLoader(pluginUrls.toArray(new URL[0]), systemLoader, allowedKlasses);
+
+    Thread.currentThread().setContextClassLoader(pluginLoader);
+
+    configDef = new ConfigDef();
+    configDef.define(configName, Type.CLASS, configNameDefault, ConfigDef.Importance.MEDIUM, "plugin class");
+
+    config = new AbstractConfig(configDef, new HashMap<>());
```
rhauch: Would it make sense to go a bit further and try getting the class from `config.getClass(configName)`, both here and above?

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
```diff
@@ -275,4 +285,79 @@ private void testValidators(Type type, ConfigDef.Validator validator, Object def
...
+
+    Class<?> systemKlass = config.getClass(configName);
+    assertEquals(systemLoader, systemKlass.getClassLoader());
+
+    // test with custom, child-first, classloader
+    Set<String> allowedKlasses = Collections.singleton(configNameDefault);
+    List<URL> pluginUrls = Collections.singletonList(Paths.get("target/classes/").toUri().toURL());
+    URLClassLoader pluginLoader = new PluginClassLoader(pluginUrls.toArray(new URL[0]), systemLoader, allowedKlasses);
+
+    Thread.currentThread().setContextClassLoader(pluginLoader);
```
gharris1727: I think you should swap back out of this loader at the end of the test (or in a finally block).

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
```diff
@@ -275,4 +285,79 @@ private void testValidators(Type type, ConfigDef.Validator validator, Object def
...
+    Class<?> pluginKlass = config.getClass(configName);
+    assertEquals(pluginLoader, pluginKlass.getClassLoader());
+    assertNotEquals(pluginKlass, systemKlass);
+    assertEquals(pluginKlass.getName(), systemKlass.getName());
+  }
+
+  // Reduced dummy implementation of a child-first classloader based on
+  // org.apache.kafka.connect.runtime.isolation.PluginClassLoader
+  // Should be used only for testing
+  public class PluginClassLoader extends URLClassLoader {
```
gharris1727: I would make this private and/or change the name to make sure that this is never confused for the real plugin class loader.
```suggestion
  private class SimplePluginClassLoader extends URLClassLoader {
```

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
kkonstantine: Not sure I follow the suggestion. 

`config.getClass(configName)` is called indeed twice. 
Once in line 303 and once in 318 (current line numbers). 

If you are referring to `configNameDefault = Password.class.getName();` probably better to not call this to leave it to the ConfigDef parsing to load this class when it constructs the defintion of the config. (Doesn't make a difference if we add this call before the definition).

rhauch:
LGTM. Thanks!

kkonstantine:
Thanks for the review @rhauch !
I replied to your comment and added a minor commit to remove the unused imports from the test (that don't break checkstyle)

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
rhauch: I missed that. Apologies.

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
```diff
@@ -275,4 +285,79 @@ private void testValidators(Type type, ConfigDef.Validator validator, Object def
...
+
+    // test with custom, child-first, classloader
+    Set<String> allowedKlasses = Collections.singleton(configNameDefault);
+    List<URL> pluginUrls = Collections.singletonList(Paths.get("target/classes/").toUri().toURL());
+    URLClassLoader pluginLoader = new PluginClassLoader(pluginUrls.toArray(new URL[0]), systemLoader, allowedKlasses);
+
+    Thread.currentThread().setContextClassLoader(pluginLoader);
+
+    configDef = new ConfigDef();
+    configDef.define(configName, Type.CLASS, configNameDefault, ConfigDef.Importance.MEDIUM, "plugin class");
```
gharris1727: I think you can re-use the same ConfigDef object, and just have the AbstractConfig constructed with the alternate classloader.

gharris1727:
Also is this fix only applicable to 5.3.x, or can it be backported farther? AK#6959 and AK#7315 were both backported to 1.0

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
kkonstantine: I construct the `AbstractConfig` in order to get the value back. My test as intended tests `parseType` inside `define`. I don't see a good reason to add a redirection here. Also, I think I would have to add the type in the map in order to call `parseType` during `AbstractConfig`'s creation. Again, given that I don't want to change visibility of methods in this PR, I'd like to keep this test as close as possible to a unit test of the parsing.

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
kkonstantine: Good idea.

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
kkonstantine: Added `protected`. Should have the same effect.

config/src/test/java/io/confluent/common/config/ConfigDefTest.java
gharris1727: Oh I see, the critical parseType is actually in `define` in the confluent ConfigDef, which is different than AK's ConfigDef. This makes my earlier suggestion invalid.

kkonstantine:
Thanks @gharris1727 
Good observation regarding backporting. The actual fixes were backported to AK 0.11.0.4
I rebased the changes after compiling against Java 7 on `3.3.x`. There a minor conflict on a single import statement in the test class which I'll resolve when merging.

kkonstantine:
Turns out that this minor conflict confuses jenkins. I'll squash and merge on the initial target branch  here (`5.3.x`) and will cherry-pick to previous branches later on.

kkonstantine:
Yeah, ticket creation is implied. Will be addressed in CC-7646
Thanks

PR #248, opened 2020-02-28
Upgrade Avro to 1.9.2

rayokota:
Upgrade Avro to 1.9.2, Jackson to 2.10.2


arvindth:
@rayokota note that the last time I tried upgrading the common pom to jackson 2.10, it caused build failures downstream in other repos like schema registry, ksql and c3. They needed some code changes to accommodate the changes in the jackson libraries.

This may have changed since I last checked, but it might be a good idea to give them a heads up.

PR #253, opened 2020-03-26
Disable dependency-check for < 5.0.x

andrewegel:
The problem:

3.0.x -> 4.1.x (inclusive) use Java7 to build (at least they should have been). Dependency-Check Maven Plugin`>=4.0.0` no longer supports Java7, so in order for maven to even be able to function, we have to use an older version (or removing it entirely is also an option). However theres anther problem: This version of dependency-check uses XML RSS Fees from NIST, and NIST has deprecated that feed in leu of JSON feeds: https://nvd.nist.gov/General/News/XML-Vulnerability-Feed-Retirement - Which this version doesn't support. At one point we were hosting these feeds in an S3 bucket, but someone deleted them, so they're effectively gone.

The solution:

Hence why I'm disabling the check altogether. Why not disable it and keep the version? The plugin *can't even load* to read that it should skip the check. I acknowledge this isn't an ideal place to be in since we're no longer scanning for CVEs, but we have a secondary check in the docker images using twistlock, so we're not 100% blind. 

I would only pint-merge this change up to 4.1.x, and `-s ours` the rest of the way.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,4 +5,5 @@ common {
   // place to trigger the rest of the pipeline. Note that we need to maintain the references to the jobs carefully since they
   // change across version branches.
   upstreamProjects = 'kafka-trunk'
+  nodeLabel = 'docker-oraclejdk7'
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,5 +5,4 @@ common {
   // place to trigger the rest of the pipeline. Note that we need to maintain the references to the jobs carefully since they
   // change across version branches.
   upstreamProjects = 'kafka-trunk'
-  nodeLabel = 'docker-oraclejdk7'
 }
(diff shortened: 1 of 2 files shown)


arvindth:
I'm not sure what's involved or what the LOE would be but have we considered converting the json feeds to xml?

andrewegel:
> I'm not sure what's involved or what the LOE would be but have we considered converting the json feeds to xml?

2-3 weeks to develop a 1-pager around an import/conversion job. Another 3-4 weeks to develop it (estimating if I did this with an AWS Lambda Function). I'd argue that this LOE would be better put towards doing maven profiles and developing a Jenkins job that runs the check with JDK8.

We have a meeting today to discuss these topics. We will run our proposed solution by your org for sign off.

andrewegel:
Closing in lue of #254

PR #257, opened 2020-04-15
ST-1323: Enable source jars if Apache or Confluent license files are found

ewencp:
This enables source jars via profiles that are activated if Apache or
Confluent license files are found. This immediately works for this
repository by including the Apache license file, but downstream
repositories using this pom as a parent need to explicitly opt-in by
having these license files, ensuring only appropriately licensed code
has source jars published.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,7 @@
         <maven-deploy.version>2.8.2</maven-deploy.version>
         <maven-enforcer-plugin.version>3.0.0-M1</maven-enforcer-plugin.version>
         <maven-jar-plugin.version>3.0.2</maven-jar-plugin.version>
+        <maven-source-plugin.version>3.2.1</maven-source-plugin.version>
         <maven-surefire-plugin.version>2.22.0</maven-surefire-plugin.version>
         <maven-site-plugin.version>3.7.1</maven-site-plugin.version>
         <maven-project-info-reports-plugin.version>3.0.0</maven-project-info-reports-plugin.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -454,6 +469,79 @@
         </plugins>
     </reporting>
     <profiles>
+        <!-- Enable source jars only if a corresponding safe license file is found - projects
```
mohnishbasha, line 472: How do we define a safe license file? Is it by detecting presence of a license file?

pom.xml
```diff
@@ -335,6 +336,20 @@
                     <artifactId>maven-deploy-plugin</artifactId>
                     <version>${maven-deploy.version}</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-source-plugin</artifactId>
+                    <version>${maven-source-plugin.version}</version>
```
mohnishbasha, line 342: Does this cover both open source and confluent owned private components?
if source jar is generated for private components, should this get promoted to public repos as well during the release?

pom.xml
```diff
@@ -454,6 +469,79 @@
...
+                        <artifactId>maven-source-plugin</artifactId>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+            <id>attach-sources-apache-parent</id>
+            <activation>
+                <file>
+                    <exists>../LICENSE-Apache</exists>
```
mohnishbasha, line 501: +1, i see that we check for if the license file exists however no necessarily check for the validity of the content of the file.

pom.xml
ewencp: Yeah, the `<file>` check is maven's mechanism for enabling the profile if the specified file exists. This set is basically just a whitelist so we're pretty conservative about when we enable this.

Ideally we'd check the license in the pom itself, but it turns out maven has a pretty limited set of ways you can enable profiles built in, and extending that requires doing special things (putting extra jars on maven classpath) to make it work, so this seems a reasonable solution given the constraints.

pom.xml
ewencp: The license file check is what protects us for non open source/source available repos. This is actually why it's been difficult to get this enabled -- finding a solution that works for that and is unlikely to accidentally publish things we're not supposed to was tricky.

NersesAM:
hey folks, any chance getting this merged?

PR #260, opened 2020-05-15
Update Scala version to 2.13

ijuma:
This aligns with Kafka and should fix some issues we've seen in muckrake for tests
that use MiniKdc.


ijuma:
It doesn't look like any of the errors are related to this change, all tests passed.

PR #264, opened 2020-06-01
ST-3461: Implement nano versioning

elismaga:
For the initial roll out of nano versioning which will only affect how we build kafka and ce-kafka all of these changes will not be made. The only changes to common will be to add the version resolver plugin we created so that we can resolve version ranges for kafka and ce-kafka, and the version properties for kafka and ce kafka will be changed to version ranges. This repo will still be built as a snapshot artifact and will not be built with a build number until we roll out nano versioning to the rest of the repos.

There are several changes in this PR that are just for testing and will be reverted in the final PR.
- Test break reporting will be re-enabled.
- The slack channel will be reverted back.
-

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,7 +23,7 @@
     <groupId>io.confluent</groupId>
     <artifactId>common-parent</artifactId>
     <packaging>pom</packaging>
-    <version>6.0.0-SNAPSHOT</version>
+    <version>6.0.0-0</version>
     <name>common</name>
     <organization>
         <name>Confluent, Inc.</name>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -296,21 +298,28 @@
                 <version>${slf4j.version}</version>
             </dependency>
 
+            <!-- Just to resolve the ce.kafka.version -->
```
ewencp, line 301: You might want a more detailed comment here about what is going on since it took quite a bit of discussion to figure all of this out.

pom.xml
```diff
@@ -696,6 +705,26 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>io.confluent</groupId>
+                <artifactId>resolver-maven-plugin</artifactId>
+                <version>1.0.0-SNAPSHOT</version>
```
ewencp, line 711: Is there a plan for pinning to a concrete version? It certainly makes sense to go through a round of testing with SNAPSHOTs first, but will there be a follow-up that pins this to a specific version?

pom.xml
```diff
@@ -781,6 +810,7 @@
             </activation>
             <properties>
                 <dependency.check.skip>false</dependency.check.skip>
+                <resolver.maven.plugin.phase>none</resolver.maven.plugin.phase>
```
ewencp, line 813: Why does the phase get set to `none` in this profile?

pom.xml
xli1996: I do want to release it, but since it will be public, I also need to publish to maven central as well, not sure if we have a process to do it?

pom.xml
elismaga: good point, will add a more detailed comment.

pom.xml
elismaga: during a CI build we run the resolver plugin via CLI so that we can get the version it resolves to and write it out to the pom file. This plugin resolves the version but doesn't update the pom file. I think @xli1996 added a proper skip property recently so I will update it to use that. I will also add a comment here describing why we skip it.

elismaga: i updated this with a comment as to why we do this.

elismaga:
i have created another PR from my fork which only implements the changes needed for rolling out nano versioning for kafka and ce-kafka so i am closing this one in favor of that one.

PR #265, opened 2020-06-05
downgrade maven-deploy-plugin to 2.8.2

xvrl:
also downgrades intall plugin to 2.5.2 since 3.0.0-M1 of install
plugin requires 3.0.0-M1 of deploy plugin according to releaes notes
https://blogs.apache.org/maven/entry/apache-maven-install-plugin-version

This is required becauses some components in our build environment are
still relying on deploy plugin 2.7.x, which prevents us from switching
to the new 3.0.x syntax for alt*DeploymentRepository which requires
removing the 'layout' component of the repo.
see https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html

PR #266, opened 2020-06-05
align plugin versions for modules not depending on common

xvrl:
This is an alternative PR to #265 and upgrades deploy and install plugins in our other modules which do not inherit from common pom. This should allow us to switch to the new syntax for `alt*DeploymentRepository` to workaround authentication issues with the deprecated syntax.

(see https://issues.apache.org/jira/browse/MDEPLOY-246)


xvrl:
@ewencp sounds good. Before we merge I'd like to have @gracechensd approve, since this will cause the build to fail once merged until we switch to the new property syntax.

PR #267, opened 2020-06-05
ST-3458: Nano versioning for kafka and ce-kafka

elismaga:
For the first phase of the nano versioning roll out we are only going to implement nano versioning for kafka and ce-kafka. I am updating the common repo to use version ranges for those two repos and adding the custom maven resolver plugin so that we can resolve those custom version ranges.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,8 +48,8 @@
         <avro.version>1.9.2</avro.version>
         <required.maven.version>3.2</required.maven.version>
         <confluent.version>6.1.0-SNAPSHOT</confluent.version>
-        <kafka.version>6.1.0-ccs-SNAPSHOT</kafka.version>
-        <ce.kafka.version>6.1.0-ce-SNAPSHOT</ce.kafka.version>
+        <kafka.version>[6.1.0-0-ccs, 6.1.1-0-ccs)</kafka.version>
+        <ce.kafka.version>[6.1.0-0-ce, 6.1.1-0-ce)</ce.kafka.version>
         <easymock.version>4.0.1</easymock.version>
         <!-- keep exec-maven-plugin on 1.5.0 until https://github.com/mojohaus/exec-maven-plugin/issues/76 is fixed
              running our LicenseFinder plugin in create-licenses-for-docker breaks when upgrading to 1.6.0 -->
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -48,8 +48,8 @@
         <avro.version>1.9.2</avro.version>
         <required.maven.version>3.2</required.maven.version>
         <confluent.version>6.0.0-SNAPSHOT</confluent.version>
-        <kafka.version>6.0.0-ccs-SNAPSHOT</kafka.version>
-        <ce.kafka.version>6.0.0-ce-SNAPSHOT</ce.kafka.version>
+        <kafka.version>[6.0.0-0-ccs, 6.0.1-0-ccs)</kafka.version>
+        <ce.kafka.version>[6.0.0-0-ce, 6.0.1-0-ce)</ce.kafka.version>
```
andrewegel, line 52: I'm going to be doing the branch cut on Monday, I have yet to see any enhancements to `release-tools` that will start replacing these values with ranges. Therefore I think I'm going to ask that you hold of shipping this PR until I've done that work on Monday, at which point, I think you'd want to apply this to a `6.0.x` branch vs master.

I'm mainly just trying to call out that theres needs to be a sync up here between @sdanduConf and myself with respect to timing, timelines, and LoE into `release-tools`.

elismaga: @andrewegel yes i will be rebasing and merging this to the 6.0.x branch. I have been coordinating the roll out of this with srini and will let you know when we have date set. As of now I just need to get the PR's reviewed before scheduling it.

pom.xml
```diff
@@ -696,6 +696,24 @@
...
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>io.confluent</groupId>
+                <artifactId>resolver-maven-plugin</artifactId>
+                <version>0.1.0</version>
+                <configuration>
+                    <groupId>org.apache.kafka</groupId>
+                    <artifactId>kafka-clients</artifactId>
+                    <versionRange>[6.0.0-0, 6.0.1-0)</versionRange>
```
elismaga, line 706: I need to make this version range point to a property so we can easily update it during branching

elismaga:
> LGTM. Only _really_ minor nit is that the commit msg says "Nanon versioning" :)

fixed

PR #268, opened 2020-06-29
ST-3458: Nano versioning for kafka and ce-kafka

elismaga:
nano versioning changes, base branch is 6.0.x

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,8 +48,8 @@
         <avro.version>1.9.2</avro.version>
         <required.maven.version>3.2</required.maven.version>
         <confluent.version>6.0.0-SNAPSHOT</confluent.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -696,6 +703,67 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
ewencp, line 706: Some commentary about what's going on here would probably help.

elismaga:
> Generally looks good. nit, but it would be good to fix the commit message that says `Nanon` instead of `Nano`

fixed the commit message

pom.xml
elismaga: @ewencp I added some comments here.

PR #274, opened 2020-07-22
ST-3988: Update jackson version to 2.10.4

niteshmor:
For jackson version 2.10.2, `jackson-dataformat-yaml` brings in snakeyaml v1.24, which is supposedly affected by CVE-2017-18640

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
         <spotbugs.maven.plugin.version>3.1.8</spotbugs.maven.plugin.version>
         <java.version>8</java.version>
         <jackson.version>2.10.4</jackson.version>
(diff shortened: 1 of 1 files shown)


niteshmor:
The pull request is against 5.4.x for 2.10.2=>2.10.4, which is only a patch update. 
5.3.x is on version 2.9.10, and going from 2.9 to 2.10 will require a bit more consideration.

niteshmor:
@confluentinc/tools separate pull requests to kafka will be required to make this change across the board.

niteshmor:
@ewencp good catch. Thanks. Updated.

niteshmor:
Looks like this didn't get applied to 5.5.x and above :( 
I thought I did a `pint merge`. Doing a manual cherry pick

PR #276, opened 2020-07-24
SEC-1305: update zk version to match that of 5.4

niteshmor:
zookeeper v3.4.14 brings in `io.netty:netty:3.10.6` which has known CVEs.
5.4.x is already on zookeeper v3.5.8 which relies on an upgraded netty version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <powermock.version>2.0.0</powermock.version>
         <slf4j.version>1.7.26</slf4j.version>
         <zkclient.version>0.10</zkclient.version>
-        <zookeeper.version>3.4.14</zookeeper.version>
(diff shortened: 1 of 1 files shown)


niteshmor:
Not sure what the FOSSA checks are doing, but Jenkins passes (including downstream validation). 
Even though it is a minor version upgrade, [zookeeper release notes](https://zookeeper.apache.org/releases.html) seem to imply that compatibility is maintained fairly well.

ijuma:
It's too risky to upgrade ZK to 3.5 in a stable branch. So, I don't think we can do this.

niteshmor:
> It's too risky to upgrade ZK to 3.5 in a stable branch. So, I don't think we can do this.

Okay, makes sense. Closing the PR

PR #283, opened 2020-08-11
ST-3461: Nano version: skip custom plugins during packaging build for archive build.

elismaga:
Do not run the custom plugins such as the resolver plugin during this build because we already resolved the kafka version ranges. I did not enable the jenkins profile which would skip these plugins because that would also change other behavior, so I am passing in these properties that cause the three plugins to get skipped. This is backwards compatible with older releases since nothing will be affected since these properties didn't exist prior to 6.1.0.

I also changed the maven command to run in batch mode so we don't get all the progress logging when downloading artifacts.

I tested this by running a packaging PR build using this branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,8 @@ ifeq ($(PULL_ARTIFACTS),yes)
 	BUILD_METHOD = newbuild
 endif
 
+MVN_SETTINGS=-Dskip.maven.resolver.plugin=true -Dcustom.install.phase=none -Dcustom.deploy.phase=none
+
 all: install
 
 archive: install
@@ -75,9 +77,9 @@ endif
 
 build:
 ifeq ($(SKIP_TESTS),yes)
-	mvn -DskipTests=true install
+	mvn -B $(MVN_SETTINGS) -DskipTests=true install
 else
-	mvn install
+	mvn -B $(MVN_SETTINGS) install
 endif
 	./create_archive.sh
 


andrewegel:
I'm concerned that this is going to be a change for *all* packaging branches, and not just master.

> I tested this by running a packaging PR build using this branch.

Please link that here. I'd also like to see this tested in 4.1.x and one of the "older" 5.x.x branches (like 5.2.x)

elismaga:
> I'm concerned that this is going to be a change for _all_ packaging branches, and not just master.
> 
> > I tested this by running a packaging PR build using this branch.
> 
> Please link that here. If you're going to go with this approach, then I would also like to see this tested in 4.1.x and one of the "older" 5.x.x branches (like 5.2.x)
> 
> An alternative approach would be to "collapse" all the archive, debian, rpm branch-bits _into_ `master` so we stop doing this funky merge-from-the `{rpm,debian,archive}` branch into the dev branch: https://github.com/confluentinc/packaging/blob/master/build/common-archive.sh#L18 before building.
> 
> This has been done for most other projects, but skipped for common and a few others.

@andrewegel What is your concern with this getting applied to the older packaging branches? It is just going to pass in some additional properties to maven that will just be ignored and never used because nothing in those older builds references them. The code that references these properties was only added on the master branch. This is why I don't feel it very necessary to test it on an older packaging build.

The PR build where I tested it was one of the builds in this PR link below, but that build still failed because of the other issue that was happening at the time in build artifacts. In the PR I just verified that the deb, archive, and rpm builds still worked and that the plugins were not run. I am having trouble finding exactly which build that was for this PR though.
https://jenkins.confluent.io/job/confluentinc-pr/job/packaging/view/change-requests/job/PR-814/

I believe there is a separate effort that was started to collapse all these branches, and that is really out of scope for this change since I need to get this fixed to have the packaging builds working post nano versioning, so I don't want to take that on in this PR.

elismaga:
@andrewegel I started another PR packaging build using these branches since I can't find the old one to link to. It may still fail though because of existing issues with packaging master. We should still be able to verify that the custom plugins didn't run though since it should happen before the ksql failure.
https://github.com/confluentinc/packaging/pull/819
https://jenkins.confluent.io/job/confluentinc-pr/job/packaging/job/PR-819/1/

elismaga:
@andrewegel I started a packaging PR build based on 5.4.x using these branches.
https://github.com/confluentinc/packaging/pull/820
https://jenkins.confluent.io/job/confluentinc-pr/job/packaging/job/PR-820/1/

andrewegel:
> @andrewegel I started a packaging PR build based on 5.4.x using these branches. confluentinc/packaging#820

Thats targeting master though. And I really just wanted a branch thats running Java7 to be tested (4.1.x) so it doesn't screw those builds up.

> This is why I don't feel it very necessary to test it on an older packaging build.

Which is why I suggested a second approach:

> An alternative approach would be to "collapse" all the archive, debian, rpm branch-bits into master so we stop doing this funky merge-from-the {rpm,debian,archive} branch into the dev branch: https://github.com/confluentinc/packaging/blob/master/build/common-archive.sh#L18 before building.

If you did that, I wouldn't feel like this change needs to be tested on older branches since it wouldn't feed into them, because the change is more targeted, and we kill some tech debt at the same time.

elismaga:
@andrewegel Now I see why you are also suggesting to collapse the branches to in order to avoid affecting the older packaging builds which I wasn't obvious to me before. I am concerned with what it would take to do that since I don't have experience doing that work yet, but would be happy to try doing that when I have some time.

Why are you concerned about testing against Java7? I don't see how this change would be affected by the version of java being used.

ewencp:
Collapsing isn't as trivial as just dumping everything into `master`, it requires reworking at least the Makefile and probably a few other things. I agree it's the best path forward and we should dedicate time to getting everything transitioned to that model, but not necessarily something I'd take on lightly in the context of a different PR.

PR #299, opened 2020-09-25
Add org.junit.jupiter testing dependencies

andrewegel:
Kafka upstream in trunk started using Junit5 [1], which refactored a lot of its components into different artifacts now. This is change adds the `org.junit.jupiter` artifacts here. Projects impacted so far:

* `common-docker` https://github.com/confluentinc/common-docker/pull/111
* `kafka-streams-examples` https://jenkins.confluent.io/job/confluentinc/job/kafka-streams-examples/job/master/2297/

[1] https://github.com/apache/kafka/pull/9231

PR #302, opened 2020-10-16
Fix incorrect junit groupid

arvindth:
https://mvnrepository.com/artifact/org.junit.vintage/junit-vintage-engine
Related to changes in #299

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -376,7 +376,7 @@
                 <scope>test</scope>
             </dependency>
             <dependency>
-                <groupId>org.junit.jupiter</groupId>
+                <groupId>org.junit.vintage</groupId>
                 <artifactId>junit-vintage-engine</artifactId>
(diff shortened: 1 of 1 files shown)


andrewegel (approved):

andrewegel:
1 question for a non-java developer: whats the difference?

arvindth:
I found this issue when fixing metadata-service's failing build. If you check my fix pr https://github.com/confluentinc/metadata-service/pull/424/files you'll see that I could skip specifying a version for `junit-jupiter-api` and it would just inherit the version from common. 

However for the incorrect artifact, I had to specify a version, since it didn't properly inherit it. Once this merges, I'll update metadata-service to remove the version for the vintage artifact also, so that one inherits from common as well.

arvindth:
Although, considering the ongoing nano versioning update, I might wait to merge this until Monday so I don't trip Eli over the weekend.

PR #309, opened 2021-01-07
SEC-1789 Bump bouncycastle

cyrusv:
Affected versions are 1.65 and 1.66
Fixed in 1.67 and 1.68, went ahead with 1.68
https://nvd.nist.gov/vuln/detail/CVE-2020-28052

This also exists only as test dep in AK, so we will merge into CP first

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,7 +86,7 @@
         <confluent-log4j.version>1.2.17-cp2</confluent-log4j.version>
         <zkclient.version>0.11</zkclient.version>
         <zookeeper.version>3.5.8</zookeeper.version>
-        <bouncycastle.version>1.66</bouncycastle.version>
+        <bouncycastle.version>1.68</bouncycastle.version>
         <checkstyle.version>8.18</checkstyle.version>
         <maven-checkstyle-plugin.version>2.17</maven-checkstyle-plugin.version>
         <maven-clean-plugin.version>3.0.0</maven-clean-plugin.version>


niteshmor:
The CVE affects only version 1.65 and 1.66. Not sure we need to backport it to branches this old. Another consideration is that we should ensure we don't have multiple versions of the same library -- one from kafka and one from common.

niteshmor (approved):
The change looks good -- you have my rubber stamp. 

BUT, before merging, we should get a sense of whether these changes will land in AK or not, and what's the timeline. And then make a plan for coordinated fixes in both places. The reason that's important is because the validation of whether we have gotten rid of the vulnerable versions or not platform wide -- that can only happen when we have *both* kafka and `common` fixed.

kkonstantine:
After generating close to 100 changelogs for connectors I have to recommend here, since it caught my eye, adding more precise and descriptive titles for targeted version upgrades. Especially if it's to address CVEs. We when generate changelogs but more importantly our customers can appreciate the info. 

Changed the title to give an example. It's quite possible I'll distill thoughts and recommendations in a wiki page some time soon.

ewencp:
@kkonstantine You actually should specifically *not* do this for CVEs as this is a public repo and may take time to generate a release that includes the fix. It's obviously a bit less sensitive with an open source dependency, but since many CVEs don't even apply, you wait until you have a release with the fix included before doing disclosure.

kkonstantine:
@ewencp are you referring to the "to address CVE" part? That's a fine but wasn't my main point. 
I was focusing on the part where we state the actual version. "Bump <dependency name>" doesn't say much to people reading the changelog.

kkonstantine:
Of course for public repos, the description is also public, and I based my comment on the fact that the CVE resolution is mentioned on the PR description already.

PR #314, opened 2021-02-18
MMA-9930: Sync Google Guava version.

xjin-Confluent:
See the last comment in https://confluentinc.atlassian.net/browse/MMA-9930
This Guava definition is already in 5.2.x and newer. 
This change will be applied to 5.0.x and 5.1.x only.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,7 @@
         <jackson.version>2.9.10</jackson.version>
         <jackson.bom.version>2.9.10.20200621</jackson.bom.version>
 
+        <guava.version>28.1-jre</guava.version>
         <jacoco-maven-plugin.version>0.8.1</jacoco-maven-plugin.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -57,6 +57,7 @@
         <jackson.version>2.9.10</jackson.version>
         <jackson.bom.version>2.9.10.20200621</jackson.bom.version>
 
+        <guava.version>28.1-jre</guava.version>
```
srpanwar-confluent, line 60: What is the version in parent poms? Can we make this change in the common pom file, so all the project can inherit the right version?

xjin-Confluent: This is the pom.xml in Common.

srpanwar-confluent: Arrggg..... stupid me :)

srpanwar-confluent (approved):

PR #325, opened 2021-04-23
ST-4106: Uplift metrics related jars to CP Images

jhlee8:
Add disk-usage-agent to common

Migrate https://github.com/confluentinc/cc-docker-base/tree/master/disk-usage-agent to `common` so we can add `disk-usage-agent.jar` as dependencies (to be accessed from artifactory) for both `common-docker/base` and `cc-docker-base` as part of the request in ST-4106.


andrewegel (commented):
I think it makes more sense to put this into https://github.com/confluentinc/common-docker as thats going to be closer to where its being utilized.

andrewegel (approved):

PR #326, opened 2021-04-26
KREST-1066 Upgrade Guava to the latest version to patch CVE-2020-8908.

dimitarndimitrov:
Downstream validation passed here, but more importantly, it also passed on `master` (see #327 and its related CI runs).
It's a bit too much to test the same on every branch between `5.2.x` and `master`. Because of that, I'll go with the assumption that no problems on the two opposite sides (`5.2.x` and `master`) is a good enough indication that there would be no problems on the in-between branches too.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
         <jackson.bom.version>2.9.10.20200621</jackson.bom.version>
 
         <gson.version>2.8.5</gson.version>
-        <guava.version>28.1-jre</guava.version>
+        <guava.version>30.1.1-jre</guava.version>
         <jacoco-maven-plugin.version>0.8.2</jacoco-maven-plugin.version>
         <junit.version>4.13</junit.version>
         <kafka.version>2.2.3-SNAPSHOT</kafka.version>


rigelbm (approved):
LGTM

dimitarndimitrov:
In addition to the downstream validation, I also did some manual validation.

I spent some time looking at the changelog for versions between `28.1` and `30.1.1` [here](https://github.com/google/guava/releases), searching for breaking changes there, and searching for the affected APIs in other Confluent repos using our internal [Hound](https://github.com/hound-search/hound) deployment.
The only breaking changes seem to be:
- net: Made MediaType reject empty type, subtype, and attribute (b080067) and non-ASCII parameter values (2278123). Also, made it serialize empty parameter values to quoted strings (2278123).
    - Some REST components use JAX-RS' `MediaType`, but Guava's `MediaType` seems unused.
- Guava types can no longer be sent over GWT-RPC. To temporarily reenable support, set the guava.gwt.emergency_reenable_rpc system property to true. (5214a10)
    - I'm not aware of any usage of GWT and GWT-RPC within Confluent.
- util.concurrent: Removed the deprecated 1-arg ServiceManager.addListener(Listener). Use the 2-arg addListener(Listener, Executor) overload, setting the executor to directExecutor() for equivalent behavior. (dfb0001)
    - ksqlDB's `ksqldb-rest-app` uses `ServiceManager`, but not its `addListener` API.

dimitarndimitrov:
Merging this change up failed immediately, as I don't have access to merge things in this repo without explicit approval. @ewencp / @elismaga / @xli1996 can you please lend me a hand here?

PR #329, opened 2021-04-28
ST-5293: Override version of netty used by ZK to address CVE

elismaga:
Exclude version of netty used by zk and use a version that has the CVE fixed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,7 @@
         <confluent-log4j.version>1.2.17-cp2</confluent-log4j.version>
         <zkclient.version>0.11</zkclient.version>
         <zookeeper.version>3.5.9</zookeeper.version>
+        <netty.version>4.1.63.Final</netty.version>
         <bouncycastle.version>1.68</bouncycastle.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -89,6 +89,7 @@
         <confluent-log4j.version>1.2.17-cp2</confluent-log4j.version>
         <zkclient.version>0.11</zkclient.version>
         <zookeeper.version>3.5.9</zookeeper.version>
+        <netty.version>4.1.61</netty.version>
```
ewencp, line 92: Do we want .61 or .63? I'm not sure how netty versioning works so not sure if those are equally safe... although I suppose this does seem like the minimal update required.

ewencp (approved):
One question about the exact version we want, but lgtm

pom.xml
elismaga: @ewencp changed the version to 4.1.63 since that is the version schema registry is already using.

ewencp: Sounds good, still lgtm

PR #332, opened 2021-05-10
Update dependency-check to 6.1.6

ijuma:
This fixes a NPE, see the following for more details:

https://github.com/jeremylong/DependencyCheck/issues/3360
https://github.com/jeremylong/DependencyCheck/issues/3361

Thanks to @dimitarndimitrov for figuring it out.


dimitarndimitrov (approved):
+1.

I think this will definitely do the trick.
I still assume that ideally we need to do this from a very early branch though (something like `5.2.x`) and for that we do need a repository admin, due to branch protection preventing `pint merge`-ing from such a branch all the way to `master`.

ijuma:
@dimitarndimitrov yeah, we should do that separately, but I want to unblock master first since this is where most development happens.

PR #334, opened 2021-05-10
Define com.fasterxml.jackson.dataformat:jackson-dataformat-cbor.

xjin-Confluent:


xjin-Confluent:
> Can you include a ticket ID at the front of the PR title and an explanation of the motivation? By itself this doesn't really do anything, and if we need to explicitly override a specific jar, I would expect there to be an addition to the `dependencyManagement` section.

Yep. Included the ticket id and also add context in the description.

ewencp:
> We could define this in the dependencyManagement in the specific repo, but that does no scale because this could be used in multiple places.

You would put it in dependencyManagement here, and then downstreams would not specify a version, they should just use the one here by default if they specify just the `group` and `artifactId` of the dependency. That's one of the values of defining things in dependencyManagement in this repo -- it makes it easy to do platform-wide upgrades of dependencies with a single change here (assuming no repos have overridden the values, which is something we keep an eye out for when reviewing PRs that add dependencies to downstream repos).

ewencp (approved):
LGTM as long as we've a) verified this against the other PR (and e.g. that there aren't compatibility issues between this version and the other jackson-bom versions) and b) this is the oldest branch we need this on.

PR #335, opened 2021-05-12
APPSEC-70: Upgrade zookeeper to 3.5.8

xjin-Confluent:
To keep consistent with https://github.com/confluentinc/common/blob/5.5.x/pom.xml


ewencp (commented):
Are there corresponding updates also to Kafka repos?

xjin-Confluent:
> Are there corresponding updates also to Kafka repos?

I will address here: https://github.com/confluentinc/common/pull/340/files

PR #336, opened 2021-05-12
CCMSG-958: Upgrade Jackson

xjin-Confluent:
Sync with versions in 5.4.x: https://github.com/confluentinc/common/blob/5.4.x/pom.xml#L58


xjin-Confluent:
Replaced by https://github.com/confluentinc/common/pull/340/files

PR #340, opened 2021-05-20
Fix zookeeper, jacksondatabind and jackson-cbor version

xjin-Confluent:
This is the process for syncing versions of libraries for June security release. This probably will break downstream build. We will coordinate teams to fix them.


xjin-Confluent:
There are downstream build failures on
 

- ksql: https://github.com/confluentinc/ksql/pull/7558
- control center: https://github.com/confluentinc/blueway/pull/2262
- schema-registry: pending

PR #342, opened 2021-05-20
Fix jacksondatabind and jackson-cbor version

xjin-Confluent:
This is to sync versions of common libraries from 5.2.x: https://github.com/confluentinc/common/commit/494caf87aa432337823c3ad2b25f554b0fd46112


xjin-Confluent:
We will fix downstream build. Schema-registry and blueway has been fixed. Pending ksql.

PR #344, opened 2021-05-21
APPSEC-110: Exclude Netty in zookeeper 3.4.14

xjin-Confluent:
Netty is only required for tls support and that was only added in ZK 3.5.x


xjin-Confluent:
Note that in common 5.4.x and later, common repo does not define netty either, so I this change will not affect downstream consumer.

xjin-Confluent:
The build failure in KSQL is not caused by this change: https://confluentinc.atlassian.net/browse/KSQL-6669

ijuma (approved):
LGTM

PR #269, opened 2020-06-29
Upgrade Jackson to 2.10.2 and Avro to 1.9.2

kkonstantine:
Upgrade jackson and jackson-bom to the most recent versions to resolve known CVEs
Upgrade Avro to 1.9.2 in 5.4.x

PR #348, opened 2021-05-25
MMA-10646: Upgrade avro version to 1.9.2

xjin-Confluent:


rayokota:
If you want to upgrade Avro, you'll have to also fix schema-registry, blueway, and ksql.  See https://github.com/confluentinc/schema-registry/pull/1228 and https://github.com/confluentinc/common/pull/269

xjin-Confluent:
Discard as we move towards another option to not upgrade avro.

PR #349, opened 2021-05-25
MMA-10646: Upgrade avro version to 1.9.2

xjin-Confluent:


xjin-Confluent:
Discard as we move towards another option to not upgrade avro.

PR #350, opened 2021-07-06
KAFKA-12756: Update ZooKeeper to v3.6.3

rondagostino:
Makes the ZooKeeper version change to 3.6.3 as was applied in https://github.com/apache/kafka/pull/10918.


ijuma:
Do we also have a metrics library issue here? Or does it only affect Kafka?

rondagostino:
> Do we also have a metrics library issue here? Or does it only affect Kafka?

 The dropwizard metrics library added in zk 3.6.3 is used by the default metrics provider within ZooKeeper -- so it is only required by the ZooKeeper server. There is no need to make it a dependency for ZooKeeper clients.

ijuma:
@rondagostino we need this in 7.0.x too, right?

rondagostino:
> we need this in 7.0.x too, right?

Yes.  I assume we would cherry-pick this to the `7.0.x` branch after merging to `master`.

ijuma:
In the non kafka repos, we merge to the older branches first and then pint merge down.

rondagostino:
Replaced by https://github.com/confluentinc/common/pull/371

PR #356, opened 2021-07-16
update mockito to 3.11.2

xvrl:
Versions 3.0.0 and later require Java 8, see https://github.com/mockito/mockito/releases/tag/v3.0.0


xli1996 (approved):
lgtm, the downstream job is unrelated to that change.

PR #270, opened 2020-07-07
SEC-1034: log4j migration to confluent repackaged version

niteshmor:
- Confluent repackaged version fixes CVE-2019-17571
- all downstream dependencies that use `slf4j-log4j12` must explicitly include the following
```
   <dependency>
       <groupId>io.confluent</groupId>
       <artifactId>confluent-log4j</artifactId>
       <version>${confluent-log4j.version}</version>
   </dependency>
```

There is a cyclic dependency in PR build success. This PR will not show a green build unless the downstream dependencies are all building fine. Those downstream dependencies will not build green unless they can get `${confluent-log4j.version}` definition from the updated `common/pom.xml`.

PR #357, opened 2021-07-22
log4j migration to confluent repackaged version

xjin-Confluent:
Apply https://github.com/confluentinc/common/pull/270/files to 5.2 and 5.3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,6 +78,7 @@
         <mockito-all.version>1.10.19</mockito-all.version>
         <os-maven-plugin.version>1.5.0.Final</os-maven-plugin.version>
         <slf4j-api.version>${slf4j.version}</slf4j-api.version>
+        <confluent-log4j.version>1.2.17-cp1</confluent-log4j.version>
         <jaxb.version>2.3.0</jaxb.version>
         <hamcrest.version>1.3</hamcrest.version>
         <!-- Update to 2.0.0 for unit test patch usage -->
@@ -155,6 +156,12 @@
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-api</artifactId>
                 <version>${slf4j.version}</version>
+            </dependency>
+            <dependency>
+                <!-- Security patches to the end-of-life log4j 1.2.17 -->
+                <groupId>io.confluent</groupId>
+                <artifactId>confluent-log4j</artifactId>
+                <version>${confluent-log4j.version}</version>
             </dependency>
             <dependency>
                 <groupId>org.apache.zookeeper</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <mockito-all.version>1.10.19</mockito-all.version>
         <os-maven-plugin.version>1.5.0.Final</os-maven-plugin.version>
         <slf4j-api.version>${slf4j.version}</slf4j-api.version>
-        <confluent-log4j.version>1.2.17-cp1</confluent-log4j.version>
+        <confluent-log4j.version>1.2.17-cp2</confluent-log4j.version>
         <jaxb.version>2.3.0</jaxb.version>
         <hamcrest.version>1.3</hamcrest.version>
         <!-- Update to 2.0.0 for unit test patch usage -->
--- a/pom.xml
+++ b/pom.xml
@@ -276,6 +276,13 @@
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-log4j12</artifactId>
                 <version>${slf4j.version}</version>
+                <exclusions>
+                    <!-- Use a repackaged version of log4j with security patches instead-->
+                    <exclusion>
+                        <groupId>log4j</groupId>
+                        <artifactId>log4j</artifactId>
+                    </exclusion>
+                </exclusions>
             </dependency>
 
             <!--this is our own artifacts, but lets others inherit-->
--- a/log4j-extensions/pom.xml
+++ b/log4j-extensions/pom.xml
@@ -35,9 +35,8 @@
 
     <dependencies>
         <dependency>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -78,6 +78,7 @@
         <mockito-all.version>1.10.19</mockito-all.version>
         <os-maven-plugin.version>1.5.0.Final</os-maven-plugin.version>
         <slf4j-api.version>${slf4j.version}</slf4j-api.version>
+        <confluent-log4j.version>1.2.17-cp1</confluent-log4j.version>
```
niteshmor, line 81: Need to use `1.2.17-cp2`. There was a followup PR that changed `-cp1` to `-cp2`

niteshmor (changes requested):
Note that this will trigger build failures in a lot of places. It might be a good idea to merge it in the mornings in the beginning/middle of the week (and not late evenings or right before a weekend).

niteshmor:
I also notice that 5.3.x is still on log4j. I am assuming the intent is to apply the same change to 5.3.x as well? @xjin-Confluent

xjin-Confluent:
> I also notice that 5.3.x is still on log4j. I am assuming the intent is to apply the same change to 5.3.x as well? @xjin-Confluent

Yes. This will be pint merged.

dimitarndimitrov:
@xjin-Confluent Shouldn't this also exclude the transitive log4j dependency from slf4j-log4j12 here: https://github.com/confluentinc/common/blob/91f91ca77ded93b8799763e789dcc51fae072b02/pom.xml#L268-L272 ?

Kafka REST (in almost all branches, including `5.2.x` and `5.3.x`) has a log4j dependency only transitively via slf4j-log4j12. - Also see these 2 previous occasions when we have been discussing this specific problem: [KREST-1029](https://confluentinc.atlassian.net/browse/KREST-1029) and [KREST-1262](https://confluentinc.atlassian.net/browse/KREST-1262).

xjin-Confluent:
> @xjin-Confluent Shouldn't this also exclude the transitive log4j dependency from slf4j-log4j12 here:
> 
> https://github.com/confluentinc/common/blob/91f91ca77ded93b8799763e789dcc51fae072b02/pom.xml#L268-L272
> 
> ?
> Kafka REST (in almost all branches, including `5.2.x` and `5.3.x`) has a log4j dependency only transitively via slf4j-log4j12. - Also see these 2 previous occasions when we have been discussing this specific problem: [KREST-1029](https://confluentinc.atlassian.net/browse/KREST-1029) and [KREST-1262](https://confluentinc.atlassian.net/browse/KREST-1262).

Nice catch. Added.

xjin-Confluent:
> Note that this will trigger build failures in a lot of places. It might be a good idea to merge it in the mornings in the beginning/middle of the week (and not late evenings or right before a weekend).

We will do announcement to teams and let them know. What do they need to do to fix the broken build?  Change their log4j into confluent-log4j? I already told teams to do so.

niteshmor (approved):
looks good

niteshmor:
> > Note that this will trigger build failures in a lot of places. It might be a good idea to merge it in the mornings in the beginning/middle of the week (and not late evenings or right before a weekend).
> 
> We will do announcement to teams and let them know. What do they need to do to fix the broken build? Change their log4j into confluent-log4j? I already told teams to do so.

If downstream projects see a build/test failure due to missing Log4j dependencies, they will have to explicitly add a new dependency on `confluent-log4j` even if they didn't have an explicit dependency on `log4j` in the past.
See an example of what the fix would look like: 
https://github.com/confluentinc/kafka-connect-elasticsearch/commit/b11285f2147422b53923a3b5f0400266a6d0f56f

dimitarndimitrov:
> > > Note that this will trigger build failures in a lot of places. It might be a good idea to merge it in the mornings in the beginning/middle of the week (and not late evenings or right before a weekend).
> > 
> > 
> > We will do announcement to teams and let them know. What do they need to do to fix the broken build? Change their log4j into confluent-log4j? I already told teams to do so.
> 
> If downstream projects see a build/test failure due to missing Log4j dependencies, they will have to explicitly add a new dependency on `confluent-log4j` even if they didn't have an explicit dependency on `log4j` in the past.
> See an example of what the fix would look like:
> [confluentinc/kafka-connect-elasticsearch@b11285f](https://github.com/confluentinc/kafka-connect-elasticsearch/commit/b11285f2147422b53923a3b5f0400266a6d0f56f)

I think the slightly bigger problem will be if some projects have a runtime dependency on Log4j (which even if rare happen exactly with things like Log4j), and these might not be obviously visible in build/test results.

xjin-Confluent:
> runtime dependency o

@niteshmor  Did we encounter any such cases when fixing 5.4.x and newer?

niteshmor:
> > runtime dependency o
> 
> @niteshmor Did we encounter any such cases when fixing 5.4.x and newer?

Runtime dependency being missing is a valid concern. It's hard to recall if we encountered any issues, it was almost an year ago. But we did do quite extensive checking and communications. We also checked various system tests to see any signs of failures because of missing classes, and looked at all CI failures in the ~24 hours period after the PR was merged. 

5.3.x and before is quite different from 5.4.x and newer, so there's a non-zero chance that you may see something different here.

xjin-Confluent:
Downstream validation errors: 
support-bundle: https://github.com/confluentinc/support-bundle/pull/37
hub-client: https://github.com/confluentinc/hub-client/pull/63
blueway: https://github.com/confluentinc/blueway/pull/2296

They should be able to get a clean build after this PR is merged since the log4j version property is defined in this PR.

xjin-Confluent:
Downstream validation will be fixed by the associated PRs. Plus, we've notified teams on potential build failures. Merge now.

PR #364, opened 2021-07-26
update checkstyle to 8.44 and fix checks

xvrl:
* update checkstyle to support recent Java versions – release notes: https://checkstyle.sourceforge.io/releasenotes.html#Release_8.44
* LineLength must now defined under Checker
* fix license comments incorrectly using javadoc style

### DIFF
--- a/build-tools/src/main/resources/checkstyle/checkstyle.xml
+++ b/build-tools/src/main/resources/checkstyle/checkstyle.xml
@@ -42,6 +42,11 @@
 
     <module name="SuppressWarningsFilter" />
 
+    <module name="LineLength">
+        <property name="max" value="100"/>
+        <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
+    </module>
+
     <module name="TreeWalker">
         <!-- Generic 'turn all checkstyle off' comment filter -->
         <module name="SuppressionCommentFilter">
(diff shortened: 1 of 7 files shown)


xli1996:
From error logs, I think those repos need to be fix first, other projects seems failed on getting dependency from those. We can notify team to fix the change.
```
rest-utils
schema-registry
ksql
```

xvrl:
@xli1996 remaining downstream failures will need to be fixes after the fact since it's not possible to be satisfy both checkstyle versions at the same time for the remaining issues.

xli1996 (approved):
Thanks for all the update PRs!

xjin-Confluent:
Are there any plan to backport this to older versions?

dimitarndimitrov:
@xvrl 
After the bump Checkstyle has started complaining about at least one of the Google code style conventions which we have mostly adopted - concise empty blocks which we very often use for one-line empty constructors.
There's a fairly wide discussion about that here: https://github.com/checkstyle/checkstyle/issues/6807 and there's ongoing work to address that.

In the meantime this has been addressed by updating the Google code style template provided by checkstyle here: https://github.com/checkstyle/checkstyle/blob/422be612a4dce54f204094449f3de75d248e2871/src/main/resources/google_checks.xml#L99-L104

As this check specifically was behind half the necessary changes in `kafka-rest`, behind almost every necessary problem in `ce-kafka-rest`, and it's not easily suppressable in `suppressions.xml` (at least I am having some problems suppressing it), I think we should implement it in the `common` checkstyle ruleset.

PR #371, opened 2021-08-03
 KAFKA-12756: Update ZooKeeper to v3.6.3

rondagostino:
Makes the ZooKeeper version change to 3.6.3 as was applied in https://github.com/apache/kafka/pull/10918.

PR #373, opened 2021-08-03
KAFKA-12756: Add dropwizard-metrics required for ZooKeeper v3.6.3

rondagostino:
ZooKeeper was upgraded to v3.6.3 in https://github.com/confluentinc/common/pull/371 but a required dependency `io.dropwizard.metrics:metrics-core` was not added this patch adds that dependency (the same version as was applied in https://github.com/apache/kafka/pull/10982, `4.1.12.1`


ijuma:
@rondagostino What changed since https://github.com/confluentinc/common/pull/350#issuecomment-874917885 ?

rondagostino:
> What changed since #350 (comment) ?


Nothing changed, actually.  It seems appropriate to explicitly define the Dropwizard metrics version here so that anyone explicitly defining a ZooKeeper dependency from here can also define a Dropwizard metrics dependency and get the correct version.  What do you think?

ijuma:
@rondagostino when would projects do this? I thought it was only needed for zk server?

rondagostino:
> when would projects do this? I thought it was only needed for zk server?

Probably never, and yes, this is only needed for ZK Server.

I thought that adding Dropwizard metrics to `<DependencyManagement>` wouldn't hurt and just makes things complete, but maybe you are correct that declaring it here might encourage misuse.

ijuma:
I think the main question is whether the platform needs to set this dependency. If it does, then it makes sense to put it here. If it doesn't, then we don't need it. If it does, we should explain under which situations. Maybe Schema Registry is one of those situations? Aside from that, I can't think why we would need it.

PR #391, opened 2021-11-19
Sync Netty Override From 6.2.x

xjin-Confluent:
This applies the same change in 6.2.x back to 5.4.x---6.1.x


xjin-Confluent:
The only build failure is spotbug  `Return value of java.util.function.BiFunction.apply(Object, Object) ignored, but method has no side effect [io.confluent.ksql.engine.rewrite.ExpressionTreeRewriterTest] At ExpressionTreeRewriterTest.java:[line 140] RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT` 5.4.x downstream failure on ksql.  When KSQL builds, this does not show up.

PR #393, opened 2021-12-04
chore: update mockito to last available 3.x version

xvrl:
Mockito 4.x has been released, updating to the latest 3.x series
so we can start removing deprecated APIs before moving to 4.x


andrewegel (approved):
PR build failed on the java11 thing. Non Blocker.

PR #400, opened 2021-12-15
Bump confluent-log4j to 1.2.17-cp6

niteshmor:


lbradstreet (approved):
LGTM, merging.

lbradstreet:
Looks like this will break downstream again but I think we have no choice but to push on with it and let all dependent projects know.

xjin-Confluent (approved):

PR #402, opened 2021-12-15
Revert confluent-log4j cp5 to cp2 for log4j concerns.

xjin-Confluent:
log4j2 2.15 also has CVE and we should wait for them to be stabilized.
log4j-cp only has log4j 1.x.
This only applies to master branch.


xjin-Confluent:
https://confluent.slack.com/archives/C02QXM7M1CY/p1639589969338700?thread_ts=1639527683.228400&cid=C02QXM7M1CY
We already have a PR to upgrade it.

PR #410, opened 2022-01-06
Revert confluent-log4j to cp2 (7.1.x only)

xjin-Confluent:
We do not plan to ship log redactor in 7.1.x of CP. 
Do not pint this to master as it is required for master branch.


cjohnson-confluent (approved):

niteshmor:
The decision to not ship logredactor in 7.1.0 is still being discussed. This should not be reverted quite yet.

xjin-Confluent:
>

xjin-Confluent:
There is no need as we will ship logredactor in 7.1.0.

PR #394, opened 2021-12-07
Add Support to Split Up Integration Tests for Junit5

michaelli321:
Junit5 uses [Tags](https://junit.org/junit5/docs/current/user-guide/#writing-tests-tagging-and-filtering) instead of Category and we can only use string instead of the IntegrationTest class that we import from here when defining integration tests. To get around this, we add a new IntegrationTest group to filter on in the maven-surefire-plugin and maven-failsafe-plugin when splitting out the integration tests

PR #420, opened 2022-01-26
Downgrade maven-surefire-plugin to 3.0.0-M4

Clarence97:
Downgrading maven-surefire-plugin back to 3.0.0-M4. The previous upgrade to 3.0.0-M5 when building with jdk11 is causing issues with all SR PR builds to fail https://jenkins.confluent.io/job/Confluent%20Public%20Repo%20PR%20builder/job/schema-registry/job/PR-2129/1/console

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,7 @@
         <maven-enforcer-plugin.version>3.0.0-M3</maven-enforcer-plugin.version>
         <maven-jar-plugin.version>3.2.0</maven-jar-plugin.version>
         <maven-source-plugin.version>3.2.1</maven-source-plugin.version>
-        <maven-surefire-plugin.version>3.0.0-M5</maven-surefire-plugin.version>
+        <maven-surefire-plugin.version>3.0.0-M4</maven-surefire-plugin.version>
         <maven-site-plugin.version>3.9.0</maven-site-plugin.version>
         <maven-project-info-reports-plugin.version>3.1.1</maven-project-info-reports-plugin.version>
         <mbknor-jackson-jsonschema.version>1.0.39</mbknor-jackson-jsonschema.version>


rayokota (approved):
LGTM

ewencp (commented):
I don't think this is the right way to fix this -- SR should figure out how to fix their builds.

* Have we verified that the original PR that this was updated for (an integration test PR for JUnit 5) still works on m4 and it was just an incidental upgrade?
* We're already *far* behind on maven-surefire-plugin versions. We will eventually need to fix the SR build anyway. Have you looked at whether newer versions fix the SR build?

Clarence97:
> I don't think this is the right way to fix this -- SR should figure out how to fix their builds.
> 
> * Have we verified that the original PR that this was updated for (an integration test PR for JUnit 5) still works on m4 and it was just an incidental upgrade?
> * We're already _far_ behind on maven-surefire-plugin versions. We will eventually need to fix the SR build anyway. Have you looked at whether newer versions fix the SR build?

@michaelli321 for the first question, is the upgrade in [your PR](https://github.com/confluentinc/common/pull/394) necessary for the functionality?

@ewencp I believe 3.0.0-M5 is already the [latest release](https://github.com/apache/maven-surefire/tags), no newer releases. This issue is only with java 11, and only happens when forkCount > 0 (default). So very likely this is the bug in the release itself. WDYT?

michaelli321:
@Clarence97 To be honest I can't quite remember. I think when I was doing [our JUnit5 upgrade](https://github.com/confluentinc/schroedinger/pull/2217) something wasn't working when trying to run both Junit4 and Junit5 tests and/or I was running into an OOM error so I upgraded to 3.0.0-M5 so that our tests aligned with the documentation (also maybe from [this StackExchange post](https://stackoverflow.com/questions/62552652/run-both-junit-4-and-junit5-with-maven-surefire-plugin-2020)?) but the errors also might have been because I was doing something incorrectly. I'm looking at [documentation](https://maven.apache.org/surefire/maven-surefire-plugin/index.html) right now and there doesn't seem to be an absolute need for 3.0.0-M5.

Since this only affects testing, I'm okay with merging the PR to see what happens and reverting if needed

Clarence97:
@michaelli321 thanks for checking. However since common is shared by multiple repos, in order not to mess up downstream tests, we'll force SR PRs to build with JDK8 for now while we confirm where that issue comes from. Closing this PR for now.

PR #429, opened 2022-02-15
import junit-bom to align junit dependencies, update to 3.8.2

xvrl:
* switch to junit-bom for junit jupiter

This simplifies the pom file and ensures downstream repositories don't
have to manually define additional junit dependencies, instead relying
on the bom to align all dependencies to the same version.

* chore: update junit jupiter to latest 3.8.2 version


andrewegel (approved):
Nevermind, I can see the failure in schema-registry-plugin in the packaging/master job, so its not related to this change.

PR #430, opened 2022-02-16
chore: update google-gson

xvrl:
Release notes: https://github.com/google/gson/blob/master/CHANGELOG.md


andrewegel (approved):
ksql failed on spot bugs in java11. Non blocker.

PR #452, opened 2022-05-06
KSECURITY-277: Fix for CVE-2022-25647

srpanwar-confluent:
Any version below 2.8.9 is vulnerable, so bumping it to 2.9.0. This will also keep it in line with what's currently in master.


xjin-Confluent (approved):

xjin-Confluent:
Blueway verification failed but oncall is working on it.

PR #453, opened 2022-05-06
Upgrade jackson bom, remove overrides

arvindth:
A quick [github search](https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Aconfluentinc+jackson.databind.version) shows that the only usage of `jackson.databind.version` is in connect repos that either override this property directly, or inherit from a connect parent pom that overrides this property. These should remain unaffected by removing these properties and dependency overrides.

Likely it's a separate effort to get connect to stop overriding these properties and dependency versions and instead inherit from common.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,10 +55,13 @@
         <spotbugs.version>3.1.12</spotbugs.version>
         <spotbugs.maven.plugin.version>3.1.12</spotbugs.maven.plugin.version>
         <java.version>8</java.version>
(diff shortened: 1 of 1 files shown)


janjwerner-confluent:
@arvindth 
This has been addressed in #533 
As you pointed out there are several connect repositories that redefine jackson.version - this is due to delay of the connectors picking up the update version of common. There is little we can do here until it gets resolved there.

PR #469, opened 2022-07-21
chore: Use https for repositories

xli1996:
# what

Start from maven 3.8.1, maven blocks external HTTP repositories by default. We should use https link for repositories

# how

Modify `http` to `https` for repositories


cyrusv (approved):

xli1996:
The build error is not related to this change.

PR #488, opened 2022-11-15
APIF-3117: Update scala version to 2.13.10.

rigelbm:


rigelbm:
See https://github.com/confluentinc/common/pull/489.

PR #502, opened 2023-02-07
fix: override maven version plugin's default versionrange

xli1996:
# what

Currently, we can not override `resolver-maven-plugin`'s default versionRange through cmd lines except changing `confluent.version.range`. However `confluent.version.range` will be used in other places to resolve versions as well.

# how

Add a new parameter `versionRange` that can be overrode through cmd lines. So that we can change `resolver-maven-plugin`'s default versionRange without affecting `confluent.version.range`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,6 +141,11 @@
             the version ranges for kafka and ce-kafka. We want to run this for local builds,
             but disable this in the jenkins profile so it doesn't run during CI builds.
         -->
+        <versionRange>${confluent.version.range}</versionRange>
+        <!-- This is the version range used by resolver-maven-plugin to resolve the
+            version of ccs and ce-kafka. This allowed this property to be overridden
+            through cmd lines.
+        -->
         <skip.maven.resolver.plugin>false</skip.maven.resolver.plugin>
         <!-- The custom install step handles installing a modified pom file with the
             kafka version ranges resolved. We only do this for local builds so for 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -141,6 +141,11 @@
             the version ranges for kafka and ce-kafka. We want to run this for local builds,
             but disable this in the jenkins profile so it doesn't run during CI builds.
         -->
+        <versionRange>${confluent.version.range}</versionRange>
+        <!-- This is the version range used by resolver-maven-plugin to resolve the
+            version of ccs and ce-kafka. This allowed this property to be overridden
+            through cmd lines.
+        -->
         <skip.maven.resolver.plugin>false</skip.maven.resolver.plugin>
```
acourtneybrown, line 149: i think you split this property from it's associated comment.

pom.xml
```diff
@@ -141,6 +141,11 @@
             the version ranges for kafka and ce-kafka. We want to run this for local builds,
             but disable this in the jenkins profile so it doesn't run during CI builds.
         -->
+        <versionRange>${confluent.version.range}</versionRange>
+        <!-- This is the version range used by resolver-maven-plugin to resolve the
+            version of ccs and ce-kafka. This allowed this property to be overridden
+            through cmd lines.
+        -->
```
acourtneybrown, line 148: can you move this comment before the right property?

acourtneybrown (commented):
couple of minor re-ordering things, then LGTM.

also, should this be against the `5.5.x` branch & then `pint merge`d to `master`.

PR #506, opened 2023-02-14
fix: remove ci.py

xli1996:
# what

This script is used to help get the correct ccs-kafka and ce-kafka versions from version range. Now as we fixed ccs-kafka and ce-kafka versions, ideally we do not need this script.

# how

Remove ci.py scripts


gracechensd (approved):

xli1996:
The downstream failure is unrelated to this change.

milosimpson (approved):
Sure

PR #510, opened 2023-02-17
fix: revert remove ci py

xli1996:
# what

This PR revert changes in https://github.com/confluentinc/common/pull/506  and un-pin nano version to fallback to use version range. The reason is that it causes confusion in release.


acourtneybrown (approved):

xli1996:
The downstream failure is unrelated to this change.

PR #517, opened 2023-04-06
APPSEC-2551: Update Jackson

janjwerner-confluent:
An update of SnakeYaml to version 2.0 revealed incompatibility with Jackson < 2.14.2. Updating Jackson to 2.14.2 that was shown working with SnakeYaml 2.0 in https://github.com/confluentinc/cc-pipeline-service/pull/1153

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,10 +55,10 @@
         <spotbugs.version>3.1.12</spotbugs.version>
         <spotbugs.maven.plugin.version>3.1.12</spotbugs.maven.plugin.version>
         <java.version>8</java.version>
-        <jackson.version>2.13.4</jackson.version>
-        <jackson.databind.version>2.13.4.2</jackson.databind.version>
-        <jackson.bom.version>2.13.4.20221013</jackson.bom.version>
-        <jackson.cbor.version>2.13.4</jackson.cbor.version>
+        <jackson.version>2.14.2</jackson.version>
+        <jackson.databind.version>2.14.2</jackson.databind.version>
+        <jackson.bom.version>2.14.2</jackson.bom.version>
+        <jackson.cbor.version>2.14.2</jackson.cbor.version>
         <gson.version>2.9.0</gson.version>
         <guava.version>30.1.1-jre</guava.version>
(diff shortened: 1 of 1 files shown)


xjin-Confluent:
This has build failure. I triggered a new build.

janjwerner-confluent:
error seemed to be unrelated. thanks for rebuilding.

xjin-Confluent (approved):

pom.xml
```diff
@@ -55,10 +55,10 @@
...
         <spotbugs.maven.plugin.version>3.1.12</spotbugs.maven.plugin.version>
         <java.version>8</java.version>
-        <jackson.version>2.13.4</jackson.version>
-        <jackson.databind.version>2.13.4.2</jackson.databind.version>
-        <jackson.bom.version>2.13.4.20221013</jackson.bom.version>
-        <jackson.cbor.version>2.13.4</jackson.cbor.version>
+        <jackson.version>2.14.2</jackson.version>
+        <jackson.databind.version>2.14.2</jackson.databind.version>
+        <jackson.bom.version>2.14.2</jackson.bom.version>
+        <jackson.cbor.version>2.14.2</jackson.cbor.version>
```
michaelli321, line 61: Why do we set all of these jackson variables instead of only using the bom's jackson version?

janjwerner-confluent: @michaelli321 
I believe this is somewhat of a technical debt - some projects had different versions of jackson sub-dependencies. You are right - the preferable solution would be to go through all the projects and use only one version. We can do such cleanup when we confirm that this jackson update does not break anything downstream.

michaelli321: I'm confused, why is this technical debt? The `jackson.bom.version` defined is the same for all other jackson versions. The parent pom should use the `jackson.bom.version` in place of the others and allow children pom's to overwrite that variable if needed

janjwerner-confluent: There are repos that use jackson variables - we should go and find all the places where those maven variables are used and normalize them to use only one. That is what I meant by the debt.

PR #553, opened 2023-10-12
Upgrade Avro

xjin-Confluent:


janjwerner-confluent (commented):
@xjin-Confluent 
Can you also add it to the dependencyManagement section. 
This way we don't have to worry about maven builds using an outdated version if they import any code from ce-kafka

xjin-Confluent:
> @xjin-Confluent Can you also add it to the dependencyManagement section. This way we don't have to worry about maven builds using an outdated version if they import any code from ce-kafka

Good idea. Added.

janjwerner-confluent (approved):
LGTM

PR #556, opened 2023-10-25
Upgrade protobuf to 3.24.4

rayokota:
Upgrade to protobuf 3.24.4  .  One of the dependencies of Schema Registry requires 3.22+, we are currently on 3.19.6


xjin-Confluent (approved):

janjwerner-confluent:
@rayokota 
was this supposed to be only for master branch?

ijuma:
We should always upgrade ce-kafka's protobuf when we upgrade common - we should add a comment saying that. We also seem to be missing some tests because this change seems to have caused a sev-1 in prod where Schema Validation is totally broken for protobuf.

PR #565, opened 2023-11-16
APPSEC-3390: update bouncycastle

janjwerner-confluent:
Update bouncycastle provider and algorithm API to match the versions used in ce-kafka.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -315,6 +315,16 @@
             </dependency>
             <dependency>
                 <groupId>org.bouncycastle</groupId>
+                <artifactId>bcpkix-jdk15on</artifactId>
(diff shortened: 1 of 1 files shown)


janjwerner-confluent:
The PR is currently failing the downstream validations. There are several downstream repositories that used bcpovjdk15 that need to be adjusted to use bcprovjdk18:
kafka-rest
rest-utils
metadata-service
schema-registry

janjwerner-confluent:
changed the PR to add a separate version of bouncycastle.jdk18.version variable to avoid downstream builds breaks

tunnelshade (approved):

PR #568, opened 2023-11-21
update zookeeper component to match version in ce-kafka

janjwerner-confluent:
In the process of mitigating Zookeeper CVE we need to update it both in ce-kafka as well as in common so the correct version is used by maven built projects.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,7 @@
         <snakeyaml.version>2.0</snakeyaml.version>
         <snappy.version>1.1.10.5</snappy.version>
         <zkclient.version>0.11</zkclient.version>
-        <zookeeper.version>3.5.9</zookeeper.version>
+        <zookeeper.version>3.8.3</zookeeper.version>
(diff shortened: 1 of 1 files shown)


xjin-Confluent (approved):

pom.xml
```diff
@@ -96,7 +96,7 @@
         <snakeyaml.version>2.0</snakeyaml.version>
         <snappy.version>1.1.10.5</snappy.version>
         <zkclient.version>0.11</zkclient.version>
-        <zookeeper.version>3.5.9</zookeeper.version>
+        <zookeeper.version>3.8.3</zookeeper.version>
```
trnguyencflt, line 99: @janjwerner-confluent what is the reason to upgrade this to 3.8.3 but not 3.7.2?

We have notice that many integration tests of kafka-rest start hanging due to the change of zookeeper version to 3.8.3 https://github.com/confluentinc/kafka-rest/pull/1221

PR #573, opened 2023-12-06
Remove kafka-connect-replicator that is removed from Jenkins

trnguyencflt:
As title said, the repo is moved to Semaphore permanently.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,6 @@
 
 common {
     slackChannel = '#kafka-warn'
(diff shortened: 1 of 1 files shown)


cchristous (approved):

trnguyencflt:
> This essentially removes the downstream check from when this repo was on jenkins. Meaning we'll no longer get an alert if a change in common breaks kafka-connect-replicator. Are we ok with that?

@arvindth I think so, kafka-connect-replicator is removed from Jenkins completely, so the check isn't applicable in Jenkins any longer. We will have that in Semaphore AFAIK.

PR #574, opened 2023-12-06
Fix CODEOWNERS

arvindth:
Previously, security would always override kafka-eng. Now, both are treated as codeowners.


janjwerner-confluent (approved):
LGTM

PR #581, opened 2024-01-29
fix: no implicit activation of docker-arm profile

lucasbru:
docker-arm profile should not be implicitly activated based on environment variables, otherwise it will break CI pipelines.


nikkobautista (approved):
Blind-approval, not sure why we're the code-owners here.

PR #595, opened 2024-04-23
[CC-26337] Version bump for Jackson Databind package

sp-gupta:


ngupta0205 (approved):
LGTM

janjwerner-confluent:
@sp-gupta @ngupta0205 
We need to revert this as this has caused several issue in the downstream validation. Please remember that common is used across a very large amount of Confluent Projects and you should make sure that downstream validation is not failing when updating dependencies in common. 
This change should have been applied in https://github.com/confluentinc/connect-plugins-common

PR #598, opened 2024-04-26
revert #595 to remediate downstream build failures

janjwerner-confluent:
Revert jackson-databind to 2.14.2 as 2.16.2 is causing downstream build failures.
Update of jackson suite needs to be coordinated with kafka / ce-kafka


sriman-confluent (approved):
LGTM

xjin-Confluent (approved):

PR #609, opened 2024-05-14
Update pom.xml

cprovencher:
bucket was migrated


janjwerner-confluent:
@cprovencher 
Could we actually remove the dependency check altogether? 
I don't believe there is any team still using that plugin.

janjwerner-confluent (approved):
LGTM, 
I believe we should remove the dependency check plugin altogether, but that can be done in next round :)

cprovencher:
@janjwerner-confluent I have no idea. If you think you have enough context to make that call, feel free to remove the plugin.

PR #616, opened 2024-05-31
unify versions of aws and azure sdks with ce-kafka

janjwerner-confluent:
This PR adds azure-identity, azure-blob-storage and aws-sdk-core to dependencymanagement to use the same versions in ce-kafka and maven build projects

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -270,8 +270,10 @@
             versions above, to match the versions in ce-kafka -->
             <dependency>
                 <groupId>com.amazonaws</groupId>
-                <artifactId>aws-java-sdk</artifactId>
+                <artifactId>aws-java-sdk-bom</artifactId>
                 <version>${aws-java-sdk.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
             </dependency>
             <dependency>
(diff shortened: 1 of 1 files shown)


ap00rv (approved):

pom.xml
```diff
@@ -263,6 +266,25 @@
...
+                <groupId>com.amazonaws</groupId>
+                <artifactId>aws-java-sdk-bom</artifactId>
+                <version>${aws-java-sdk.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>com.azure</groupId>
+                <artifactId>azure-identity</artifactId>
+                <version>${azure-identity.version}</version>
```
xjin-Confluent, line 281: Should we also place a scope <scope>import</scope> for azure?

janjwerner-confluent: no, this is regular dependency, only bom typed dependencies should be in scope import

janjwerner-confluent: see line 375/397 for netty-bom

xjin-Confluent (approved):

PR #619, opened 2024-06-12
Remove dependency-check-maven plugin

cchristous:
This removes the `dependency-check-maven` plugin from `pom.xml`. This plugin was once upon a time used by AppSec supposedly, but it is no longer used.


janjwerner-confluent (approved):
LGTM, Thanks Corey!

PR #622, opened 2024-06-18
APPSEC-4267: update azure-identity to 1.12.2

janjwerner-confluent:
Update azure identity to resolve CVE-2024-35255

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,8 +48,8 @@
         <argLine></argLine>
         <avro.version>1.11.3</avro.version>
         <aws-java-sdk.version>1.12.701</aws-java-sdk.version>
-        <azure-identity.version>1.12.2</azure-identity.version>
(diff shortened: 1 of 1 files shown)


trnguyencflt:
@janjwerner-confluent could you please check if this version of azure-identity match the corresponding version of azure-storage in azure parent pom. The two dependencies need to be upgrade together. In the past, we had an issue where only one azure-identity is updated, which caused issue in storage module, which uses azure-storage

janjwerner-confluent:
Hey Truc, 
Thanks for checking, we can wait with this update until the end of the month (we have code freeze anyways) and update all the azure dependencies in unison. There is KSECURITY-4212 that handles this issue for ce-kafka.

trnguyencflt:
> Thanks for checking, we can wait with this update until the end of the month (we have code freeze anyways) and update all the azure dependencies in unison

make sense, thanks 👍

PR #702, opened 2024-10-22
add okio-jvm to dependencyManagement to avoid regressions from ce-kafka

janjwerner-confluent:
Add okio-jvm to dependencyManagement section as it is brought in as a transitive dependency of kafka-client / apiserver code introduces in ce-flink


flabbergastedbd (approved):

smjn (approved):
LGTM

PR #703, opened 2024-10-22
add okio-jvm to dependencyManagement to match ce-kafka

janjwerner-confluent:
Add okio-jvm to avoid regressions introduced by adding ce-flink apiserver-client in ce-kafka.


flabbergastedbd (approved):

smjn (approved):
LGTM

PR #739, opened 2024-11-29
upgrade sem java version to 17 as we already upgraded to 17 here.

princepurohit:
we already upgraded here:
https://github.com/confluentinc/common/commit/34f3ad497b12b0316e24116a275f3aca56bb23bc


janjwerner-confluent (approved):
lgtm

PR #741, opened 2024-12-02
Revert "upgrade sem java version to 17 as we already upgraded to 17 h…

princepurohit:
…ere."

This reverts commit 4b2f8e9bf3f979f8a8b029c1f88247696ef08279.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -76,7 +76,7 @@ after_pipeline:
       - name: SonarQube
         commands:
           - checkout
-          - sem-version java 17
+          - sem-version java 11
           - artifact pull workflow target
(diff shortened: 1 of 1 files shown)


princepurohit:
I dont think it resolved original issue and hence reverting to avoid any side effects/surprises.
https://github.com/confluentinc/common/commit/4b2f8e9bf3f979f8a8b029c1f88247696ef08279

janjwerner-confluent (approved):
lgtm

princepurohit:
issue was cp build failing due to version bump but I feel we need better branching strategy.
cp uses different branches for most projects.
however there is some nuance due to which connect team had preferred to use latest branch and hence it might have lead to issue.
https://confluent.slack.com/archives/CLS8T8K4K/p1732874846391109

PR #759, opened 2025-02-13
Upgrade Protobuf to 4.29.3

rayokota:


janjwerner-confluent (approved):
LGTM, 
This is a major version upgrade. This should be backwards compatible, see e.g., https://github.com/protocolbuffers/protobuf/blob/v3.0.0-beta-4/CHANGES.txt
there are breaking changes for non-java 
https://github.com/protocolbuffers/protobuf/releases
I believe other major component owners should chime in on it. Let's get the ball rolling on testing this and revert if needed.

PR #729, opened 2024-11-11
update java to version 17 and avro to 1.120

janjwerner-confluent:
Merging together #707 and #721

PR #764, opened 2025-03-05
[Do not merge] test downgrade of Java version to 11

janjwerner-confluent:
Folks, 
This PR is to discuss the java version used across CP 8.0.
In https://github.com/confluentinc/common/pull/729
we have updated Java to version 17 per request from CPBR. Unfortunately, per @rayokota this is causing problems with clients expecting Java 11. 
I expect we'll need to delineate versions at some place - I am wondering what clients explicitly need Java 11 support? 
Should we set minimum version of Java to 11 in common and 17 in rest-utils?


rayokota:
Closing in favor of https://github.com/confluentinc/common/pull/776

PR #777, opened 2025-04-16
Use JDK 11 for client compatibilty

rayokota:
We have updated Java to version 17 per request from CPBR. Unfortunately, this is causing problems with clients expecting Java 11. 
We'll explicitly set to 17 in rest-utils, ksql, and other repos that run server-side


ijuma (approved):
LGTM, thanks.

PR #782, opened 2025-04-21
remove zookeeper license / notice

janjwerner-confluent:
CP 8.0 ships without Zookeeper, we should remove the notices accordingly


rayokota (approved):
Thanks, lgtm

PR #885, opened 2025-12-18
MINOR: Update jackson versions to match kafka versions

omkreddy:
jackson-annotations follows a different versioning scheme starting from 2.20, so I’ve added a new property, `jackson.annotations.version`, for the jackson-annotations library. This property needs to be used by downstream projects.

https://github.com/confluentinc/ce-kafka/pull/26252

We will update downstream projects after this

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,8 @@
         <spotbugs.version>4.9.3</spotbugs.version>
         <spotbugs.maven.plugin.version>4.9.3.0</spotbugs.maven.plugin.version>
         <java.version>11</java.version>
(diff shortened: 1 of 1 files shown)


aliehsaeedii (approved):

janjwerner-confluent:
@omkreddy 
Could we consider reverting this? 
the jackson-annotations version is already defined in jackson bom ( we import that). I think that the better way to solve this issue in rest utils would be to remove redefinition of jackson versions instead of adding this property. 
I will add it to 8.1.x and 8.2.x ( we have failing builds there after jackson update), but I think the cleaner solution is to remove it altogether.

PR #1044, opened 2026-07-14
Remove properties for deps fully managed by confluent-common-bom

nlou9:
## Summary
- Removes 8 `<properties>` entries (`classgraph.version`, `commons-codec.version`, `commons-compress.version`, `commons-validator.version`, `grpc.version`, `jose4j.version`, `okio.version`, `snakeyaml.version`) that existed only for downstream backward-compat.
- Per the [DP-18877](https://confluentinc.atlassian.net/browse/DP-18877) downstream-usage audit, none of these have a live inheritor: repos that use these artifacts already override the version locally, and each value exactly matches what `confluent-common-bom` (already imported here) manages — so removal is a no-op for every consumer.
- `gson.version` is intentionally **not** touched in this PR — `schema-registry` still inherits it directly and it has drifted from the BOM's version (2.9.0 vs 2.11.0), so it needs downstream migration first (tracked via https://github.com/confluentinc/schema-registry/pull/4423).

## Test plan
- [ ] CI build passes
- [ ] Confirm no downstream repo in the common-parent inheritance chain still references any of the removed properties

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[DP-18877]: https://confluentinc.atlassian.net/browse/DP-18877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1046, opened 2026-07-15
[URGENT] Restore jose4j.version property (regression from #1044)

nlou9:
## Summary
- **Regression:** PR #1044 removed `jose4j.version` from `common`'s properties, assuming nothing referenced the inherited value directly. That assumption was wrong.
- `confluent-security-plugins-common`'s `pom.xml` declares an explicit `<dependency>` on `org.bitbucket.b_c:jose4j` versioned via `${jose4j.version}` inherited from `common-parent`. Without the property, Maven fails at **POM-parsing time**:
  ```
  'dependencies.dependency.version' for org.bitbucket.b_c:jose4j:jar must be a valid version but is '${jose4j.version}'
  ```
- Confirmed via CP Jar Build CI Gating on [PR #1045](https://github.com/confluentinc/common/pull/1045) — `Build security jar` fails with this exact error. Since #1044 is already merged into `8.0.x`, this is a **live regression**, not just a risk.
- This restores only the property (not the `dependencyManagement` entry, which correctly stays removed — `confluent-common-bom` already manages `jose4j` transitively).
- Companion fix in progress: a PR against `confluent-security-plugins` (8.0.x) to drop its explicit `${jose4j.version}` reference, so this property can be safely removed from `common` again once that lands and is confirmed safe.

## Test plan
- [ ] CI passes, including CP Jar Build CI Gating (specifically `Build security jar`)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,7 @@
         <jakarta-websocket-api.version>2.2.0</jakarta-websocket-api.version>
         <jakarta-servlet-api.version>6.1.0</jakarta-servlet-api.version>
         <jetty.version>12.0.36</jetty.version>
+        <jose4j.version>0.9.6</jose4j.version>
         <gson.version>2.11.0</gson.version>
         <guava.version>32.0.1-jre</guava.version>
         <jacoco-maven-plugin.version>0.8.13</jacoco-maven-plugin.version>


jolshan:
Do you still want this review or should I wait until it is no longer a draft?

jolshan (approved):

nlou9:
> Do you still want this review or should I wait until it is no longer a draft?

Thanks for reviewing.

PR #1072, opened 2026-08-17
Add tools.jackson.databind to common pom

janjwerner-confluent:
Add managed version and dependencyManagement section for tools.jackson.databind - transitive dependency used in Schema Registry [JSONata4Java](https://mvnrepository.com/artifact/com.ibm.jsonata4java/JSONata4Java)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,7 @@
         <java.version>11</java.version>
         <jackson.version>2.22.1</jackson.version>
         <jackson.annotations.version>2.22</jackson.annotations.version>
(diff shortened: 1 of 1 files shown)


janjwerner-confluent:
@nlou9 
Adding this dependency here - it shouls be likely removed after update in schema registry / and / or moved to Confluent BOM

shayelkin (approved):
We should revert https://github.com/confluentinc/ce-kafka/pull/34008 once this gets applied to the SR code we're bringing in.

PR #1083, opened 2026-08-25
Remove unused httpclient5.version and gson.version properties

nlou9:
## Summary
- Drop `httpclient5.version` (5.6.4) and `gson.version` (2.11.0) from the root `pom.xml` properties.
- Both are already managed by `confluent-common-bom` 0.1.9 (current on this branch) at the identical values, and `common` has no local `dependencyManagement` entry for either — they were pure leftover duplicates. No BOM bump needed for this change.
- Verified no downstream consumer inherits either property: known consumers of `gson.version` (`bamboo`, `confluent-cloud-plugins`, `kafka-connect-elasticsearch`) and `httpclient5.version` (`schroedinger`) all pin their own local value.

## Test plan
- [ ] CI build passes on this branch
- [ ] `mvn help:effective-pom` confirms `gson`/`httpclient5` still resolve (from the BOM) at the same versions as before

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,6 @@
         <azure-identity.version>1.17.0</azure-identity.version>
         <azure-storage.version>12.31.2</azure-storage.version>
         <azure-security-keyvault-keys.version>4.10.7</azure-security-keyvault-keys.version>
-        <httpclient5.version>5.6.4</httpclient5.version>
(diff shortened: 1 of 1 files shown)


ChalonYang (approved):

janjwerner-confluent:
LGTM, Thanks @nlou9

PR #1090, opened 2026-08-27
Bump confluent-common-bom to 0.1.13

nlou9:
## Summary
- Bumps `confluent-common-bom.version` 0.1.10 -> 0.1.13.
- 0.1.13's only change vs 0.1.12 is a revert of `spiffe.version` 0.8.17 -> 0.8.14 (confluentinc/confluent-common-bom#69), fixing a `NoSuchMethodError` at runtime against `io.confluent.srj:spire` (which is compiled against 0.8.14 and hadn't been recompiled against the newer, binary-incompatible `java-spiffe-core`).
- Confirmed `common` does not declare any dependency on `io.spiffe:*` artifacts anywhere in the repo (root `pom.xml` or any module `pom.xml`) — grepped the whole tree for `spiffe`, no hits. So this bump is otherwise a no-op for `common` besides the version number itself; there was no local spiffe pin to change.

## Verification
- `mvn validate` passes.
- `mvn help:effective-pom` confirms `io.spiffe:java-spiffe-provider`/`java-spiffe-core`/`grpc-netty-linux` all resolve to `0.8.14` via the BOM's dependency management (not that any module actually depends on them — `common` doesn't).
- No network access to Confluent's internal artifact repo from this environment to do a full `dependency:tree` resolution check beyond `effective-pom`; flagging per the standard verification note.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
         <argLine></argLine>
         <avro.version>1.12.2</avro.version>
         <commons-io.version>2.20.0</commons-io.version>
-        <confluent-common-bom.version>0.1.12</confluent-common-bom.version>
+        <confluent-common-bom.version>0.1.13</confluent-common-bom.version>
         <aws-java-sdk.version>1.12.746</aws-java-sdk.version>
         <aws-java-sdk-v2.version>2.29.39</aws-java-sdk-v2.version>
         <azure-identity.version>1.17.0</azure-identity.version>
--- a/pom.xml
+++ b/pom.xml
@@ -48,8 +48,7 @@
         <argLine></argLine>
         <avro.version>1.12.2</avro.version>
         <commons-io.version>2.20.0</commons-io.version>
-        <confluent-common-bom.version>0.1.11</confluent-common-bom.version>
-        <commons-beanutils.version>1.11.0</commons-beanutils.version>
+        <confluent-common-bom.version>0.1.12</confluent-common-bom.version>
         <aws-java-sdk.version>1.12.746</aws-java-sdk.version>
         <aws-java-sdk-v2.version>2.29.39</aws-java-sdk-v2.version>
         <azure-identity.version>1.17.0</azure-identity.version>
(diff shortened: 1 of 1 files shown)


nlou9:
Correction: the \`mockito.version\` bump (4.6.1 → 5.23.0) in the earlier commit has been reverted.

Semaphore CI caught it via a downstream build ([job 2609cbb7](https://semaphore.ci.confluent.io/jobs/2609cbb7-c4ee-49de-9345-b61a542b194d), packaging repo, "Build rest-utils jar"): \`rest-utils\` declares both \`mockito-core\` and \`mockito-inline\` at \`${mockito.version}\`, inherited from \`common\` via \`common-parent\`. \`org.mockito:mockito-inline\` was discontinued after \`5.2.0\` (Mockito folded inline mocking into \`mockito-core\` by default for the 5.x line), so bumping the shared property broke \`rest-utils\`'s dependency resolution even though nothing in \`common\` itself uses \`mockito-inline\`.

\`common\`'s own \`mockito-core\`/\`mockito-junit-jupiter\` still resolve to \`5.23.0\` via confluent-common-bom's own dependencyManagement entries regardless of this property's value, so reverting it has no effect there — it only stops driving the now-broken \`mockito-inline\` resolution for inheritors like \`rest-utils\`.

pom.xml
```diff
@@ -247,17 +246,6 @@
...
-              <groupId>org.apache.avro</groupId>
-              <artifactId>avro</artifactId>
-              <version>${avro.version}</version>
-            </dependency>
-            <!-- Pin version of commons-beanutils as it is used transitively not only by commons-validator -->
-            <dependency>
-                <groupId>commons-beanutils</groupId>
-                <artifactId>commons-beanutils</artifactId>
-                <version>${commons-beanutils.version}</version>
-            </dependency>
```
ChalonYang, line 260: it seems these are in master branch? https://github.com/confluentinc/common/blob/8c19f4310c99dd091528e0a8fb6b2bb628348066/pom.xml#L251-L262

why are we removing in 8.0.x?

pom.xml
```diff
@@ -46,10 +46,9 @@
     <properties>
         <!--jacoco prepare-agent needs argLine defined to set this variable. if we don't have a default set then maven gets angry.-->
         <argLine></argLine>
-        <avro.version>1.12.1</avro.version>
+        <avro.version>1.12.2</avro.version>
         <commons-io.version>2.20.0</commons-io.version>
-        <confluent-common-bom.version>0.1.10</confluent-common-bom.version>
```
ChalonYang, line 51: The PR summary says:
> Bumps confluent-common-bom.version 0.1.12 -> 0.1.13.

but actually from 0.1.10 -> 0.1.13, could you update?

ChalonYang (commented):
Thanks for the PR, left 2 comments

pom.xml
nlou9: we will remove these from 8.0.x and pint merge to master branch

pom.xml
nlou9: updated! thanks for pointing out the difference.

ChalonYang (approved):

PR #1102, opened 2026-09-03
Remove unused easymock.version pin

nlou9:
## Summary
- Removes the `easymock.version` property and its `org.easymock:easymock` `dependencyManagement` entry from `common`'s root `pom.xml`.
- `confluent-common-bom` 0.1.10 (already imported here) manages `org.easymock:easymock` at the same `5.6.0` version, so this pin is redundant — same resolved version either way.
- No repo on the 8.x/master release train inherits `easymock.version` from this branch of `common`. The one downstream reference found (`kafka-connect-storage-cloud`) resolves through an unrelated, pre-BOM `common` release (`7.7.x`) via its own parent chain (`kafka-connect-storage-common-parent`), so removing this from `8.0.x` doesn't affect it.

## Verification
- `mvn -N validate` passes
- No `.java` or `pom.xml` file in this repo references `easymock`/`EasyMock` directly — the entry only existed to supply downstream consumers, which the BOM already covers.

## Tracking
Part of the BOM-migration cleanup identified in the [Common Inheritance Map](https://claude.ai/code/artifact/854856e4-47a3-4455-9e22-326715a12c82) audit — `easymock.version` was the property with the fewest downstream repos (1).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,6 @@
         <required.maven.version>3.2</required.maven.version>
         <kafka.version>[8.0.8-0-ccs, 8.0.9-0-ccs)</kafka.version>
         <ce.kafka.version>[8.0.8-0-ce, 8.0.9-0-ce)</ce.kafka.version>
-        <easymock.version>5.6.0</easymock.version>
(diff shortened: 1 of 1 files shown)


kpatelatwork (approved):
LGTM

PR #1127, opened 2026-09-15
Migrate to azure-sdk-bom instead of pinning Azure SDK artifacts individually

nlou9:
## Summary
- Replaces `azure-identity.version`, `azure-storage.version`, and `azure-security-keyvault-keys.version` (each independently pinned) with a single `com.azure:azure-sdk-bom` import at `1.3.8`, matching the existing pattern already used here for `aws-java-sdk-bom` and `software.amazon.awssdk:bom`.
- The comment above these deps claims the versions are pinned "to match the versions in ce-kafka", but that had drifted: `ce-kafka`'s `8.0.x` branch (`dependencies/dependencies.bzl`) pins `azure-identity 1.18.4` and `azure-storage-blob 12.35.0`, both newer than what was pinned here (`1.17.0` / `12.31.2`). `azure-sdk-bom:1.3.8` resolves to exactly `ce-kafka` 8.0.x's pinned versions (`azure-core 1.58.1`, `azure-identity 1.18.4`, `azure-storage-blob 12.35.0`, `azure-security-keyvault-keys 4.11.1`) — this also matches `ce-kafka` master's own `azure-sdk-bom:1.3.8` import.
- Verified no downstream repo depends on the removed properties via Maven inheritance: an org-wide code search found every other repo referencing `azure-identity.version`/`azure-storage.version` (`kafka-connect-azure-event-hubs`, `kafka-connect-azure-blob-storage`, `OpenMetadata`, `kafka-connect-object-store-source`) defines its own local copy of the property, which shadows rather than inherits from `common`'s definition.

## Test plan
- [x] `mvn -N validate` — pom is well-formed.
- [x] `mvn -N help:effective-pom`: confirmed resolved `dependencyManagement` versions are exactly `azure-core:1.58.1`, `azure-identity:1.18.4`, `azure-storage-blob:12.35.0`, `azure-security-keyvault-keys:4.11.1`.


JordanSekky (approved):

janjwerner-confluent (approved):
Looks good to me.

